Issue #5, opened 2021-06-02
Setup the frontend codebase for the dashboard


ericrfdev:
Setup the following tech stack for the frontend:
- **UI framework**: React with Typescript
- **Syling**: Tailwind and TailwindUI
- **Routing**: React Router DOM
- **Hosting**: Netlify

The rest of the libraries will be installed as we need them.

Additionally to avoid issues, we'll use **[Nodenv](https://github.com/nodenv/nodenv)** to lock the node version to the current LTS. (**14.17.0** at the time of writing).

PR #6, opened 2021-06-02
(#5) Setup the frontend codebase for the dashboard

ericrfdev:
- Install React with Typescript through create-react-app
  - Install React Router DOM
  - Install Tailwind, TailwindUI and it's required dependencies (PostCSS, autoprefixer, @headlessui/react, @heroicons/react)
  - Setup Inter font as default one. (TailwindUI components have been created using this font)
  - Use CRACO to override the default PostCSS config from create-react-app. This is needed because Tailwind it's a PostCSS plugin and needs to be listed in it's config file
  - Lock the node version to the current LTS (14.17.0)
  - Install prettier and setup a minimum conf. I decided to not setup a precommit hook at this stage
  - Add Netlify config file


nandanrao (approved):
Looks reasonable to me!

Issue #8, opened 2021-06-07
Create initial types for the Study data that will be displayed in the app


ericrfdev:
The purpose of these types is to:
- Make explicit the Study data that is required to display the different tables and dashboards in the UI.
- Uncover issues/doubts
- Solve existent doubts

The Study data that the types are trying to model, can be found here: https://github.com/vlab-research/vlab/issues/3

PR #9, opened 2021-06-07
(#8) Create initial types for Study data

ericrfdev:


dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+interface StudyProgress {
+  currentParticipants: number;
+  expectedParticipants: number;
+  currentAverageDeviation: number;
+  expectedAverageDeviation: number;
+  date: string;
+  averageDeviationOverTime: DataPoint[];
+}
+
+interface Strata {
```
nandanrao, line 19: I would call this `StratumProgress`, in line with `StudyProgress`.

dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+  id: string;
+  percentageDesired: number;
+  currentPercentage: number;
+  remainingPercentage: number;
+  expectedPercentage: number;
+  currentPriceToAdquireParticipant: number;
```
nandanrao, line 25: "acquire" in english - no "d" ;-

But I actually think it's unnecessarily verbose and not entirely accurate. To some degree, if there were to be multiple prices per participant, "price to acquire" could be the price to get them to start a survey and here we really have the "price per completed survey"...

Options I like better: 

1. `currentParticipantPrice`
2. `currentPricePerParticipant`

dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+  currentBudgetDecided: number;
```
nandanrao, line 26: I would prefer "set" or "configured". It's not just that it decided the budget, it actually set it on facebook!

dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+  remainingPercentageOverTime: DataPoint[];
```
nandanrao, line 27: I like the idea of having this, but I don't like nesting this. 

`StratumProgress` should have a datetime. 

Then you can have an array of `StratumProgress` which has _all_ values over time (you can select `remainingPercentage` easily if wanted to plot, or anything else!

nandanrao (changes requested):
Overall, great idea to start with the types, it's good for discussion and nailing down naming conventions!

dashboard-frontend/src/types/study.ts
nandanrao: Ah, the same should apply for `StudyProgress` actually!

nandanrao: (remove averageDeviationOverTime and instead anything should be able to be created "overTime" - it's just a view on any `StudyProgress[]`.

dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+  currentAverageDeviation: number;
+  expectedAverageDeviation: number;
+  date: string;
+  averageDeviationOverTime: DataPoint[];
+}
+
+interface Strata {
+  id: string;
+  percentageDesired: number;
+  currentPercentage: number;
```
nandanrao, line 22: Ah, in addition to the `percentage` numbers, we need the `absolute` numbers too. 

The former could be computed from the latter on the client-side, of course. But we need it everywhere, so maybe it's better to just keep shuffling both types around everywhere.

dashboard-frontend/src/types/study.ts
```diff
@@ -0,0 +1,33 @@
...
+  remainingPercentage: number;
```
nandanrao, line 23: In `StudyProgress` we're calling this "devation" and here we're calling it "remaining" -- we should pick one word. 

Probably, "deviation" is better because it can be above or below!

dashboard-frontend/src/types/study.ts
ericrfdev: Yes, I like it 👍

dashboard-frontend/src/types/study.ts
ericrfdev: Yes your right about this, it can be misleading if we have multiple prices.

Both options you proposed look good to me. I will go for **currentPricePerParticipant**,  I think it's slightly more clear than option 1.

dashboard-frontend/src/types/study.ts
ericrfdev: Yes, makes sense 👍

Now that **StratumProgress** is a point/set of data in a specific datetime, what do you think about just calling it **currentBudget** ?

dashboard-frontend/src/types/study.ts
ericrfdev: Yes I will update the code 👍

Actually it crossed my mind to promote **StratumProgress** (before called Strata) into a data point.

But I was not sure about if we could aggregate all the data from an **StratumProgress** at an specific point in time.

I knew that we could for **remainingPercentage** because of this dashboard -> https://user-images.githubusercontent.com/7491490/119506873-9d6ab380-bd6e-11eb-8fbf-c2ff7caae8bc.png 
and that's why I decided to separe it.

About the **StudyProgress**, the same reasoning I know that we can aggregate the current average deviation at a very granular time intervals but was not sure about the other values.

dashboard-frontend/src/types/study.ts
ericrfdev: I renamed it to **percentageDeviation**

dashboard-frontend/src/types/study.ts
ericrfdev: Makes sense!

From having:
```
percentageDesired: number;
currentPercentage: number;
```

Now, we have:
```
desiredPercentage: number;
currentPercentage: number;
desiredValue: number;
currentValue: number;
```

I agree on what you say about not having to compute some things on the frontend. I added a **name field** to the **StratumProgress** interface.

I think that is good idea to treat id as an opaque token independently that right now the id is actually the name, what do you think on having this extra **name field** ?

dashboard-frontend/src/types/study.ts
ericrfdev: In the **StudyProgress** interface I also added an **id field**.

The frontend needs it in order to diff re-renders etc..

From the frontend side the **id** will be an opaque token (that is unique), implementation wise in the api we can reuse datetime or what we find more appropiate when we implement it.

dashboard-frontend/src/types/study.ts
```diff
@@ -3,31 +3,29 @@ export interface Study {
...
-  currentBudgetDecided: number;
-  remainingPercentageOverTime: DataPoint[];
-}
-
-interface DataPoint {
-  date: string;
-  value: number;
-}
+  percentageDeviationFromGoal: number;
+  desiredValue: number;
```
nandanrao, line 27: Why not have this called `currentParticipants` just like in StudyProgress? It's the same value, just per-stratum... Same with `desiredParticipants`...?

dashboard-frontend/src/types/study.ts
nandanrao: Yeah of course, sounds reasonable... Right now there is nothing enforced for uniqueness at DB level, but we can change that. A report should be unique. Maybe will add UUID at DB level... Will think about it...

dashboard-frontend/src/types/study.ts
ericrfdev: I thought about it but I ended up using generic names because Stratum seems generic to me.

But you're right it's 0 beneficial to make this generic in the context of an Study. I'll upload the changes now!

dashboard-frontend/src/types/study.ts
nandanrao: I agree, it's clear!

ericrfdev:
@nandanrao When you give the approval, wait to merge it.

I'll squash all the commits to remove noise from the git history. I'm only adding them to make the review time easier.

nandanrao:
I see - sounds good!

Another workflow you can just make a feature branch and we merge into feature branch with PR's and then you squash feature branch commits before merging into master.

nandanrao:
But yeah, looks good to me!

Issue #13, opened 2021-06-08
Create a generator for fake Study data


ericrfdev:
The task will allows us to:
- Implement the whole UI without the need of an api that provides real data
- Implement the functionality of refreshing the data at different refresh rates in all the dashboards and tables
- Become more familiar with the underlying domain and how the different data is interconnected
- Discover limitations, misunderstandings and be sure that everyone is in the same page

PR #21, opened 2021-06-13
(#13) Create a Builder for generating fake Study data

ericrfdev:


dashboard-frontend/src/types/study.ts
```diff
@@ -1,31 +1,36 @@
+import { Currency } from './shared';
+
 export interface Study {
   id: string;
   name: string;
   slug: string;
-  createdAt: string;
+  createdAt: number;
```
ericrfdev, line 7: @nandanrao 
I changed this **createdAt** and the rest of the **datetime's** to be **timestamps** instead of using **ISO 8601**.

pd: I found this article https://www.moesif.com/blog/technical/timestamp/manage-datetime-timestamp-timezones-in-api/ that touch on the topic and it's a nice read 👍

dashboard-frontend/src/types/study.ts
```diff
@@ -1,31 +1,36 @@
...
   expectedPercentage: number;
   percentageDeviationFromGoal: number;
   desiredParticipants: number;
+  expectedParticipants: number;
   currentParticipants: number;
   currentBudget: number;
   currentPricePerParticipant: number;
-}
\ No newline at end of file
+  currency: Currency;
```
ericrfdev, line 35: @nandanrao Not sure if at the moment we are only dealing with **usd** but I added **currency** since I imagine that we want to have support for other ones.

dashboard-frontend/src/types/study.ts
```diff
@@ -1,31 +1,36 @@
...
   studyProgressList: StudyProgress[];
   stratumProgressList: StratumProgress[];
 }
 
-interface StudyProgress {
+export interface StudyProgress {
   id: string;
-  datetime: string;
+  datetime: number;
+  desiredParticipants: number;
```
ericrfdev, line 15: I think this one was missing here. In StratumProgress we have it.

dashboard-frontend/src/types/study.ts
```diff
@@ -1,31 +1,36 @@
...
   id: string;
   name: string;
-  datetime: string;
+  datetime: number;
   desiredPercentage: number;
   currentPercentage: number;
   expectedPercentage: number;
   percentageDeviationFromGoal: number;
   desiredParticipants: number;
+  expectedParticipants: number;
```
ericrfdev, line 31: @nandanrao This one was also missing

dashboard-frontend/src/types/study.ts
nandanrao: Hrm, seems like the blog says "use either", which seems reasonable to me, but yeah go for timestamps I'm fine with that!

dashboard-frontend/src/types/study.ts
nandanrao: Hrm, yes, but I'm actually not sure if we ever need to know it. Or will know it. Their Facebook Ad Account will be set to the currency and we just deal with numbers...

dashboard-frontend/src/types/study.ts
nandanrao: Interesting. Yeah let's have it. It won't be there for every study, nor will it be in StratumProgress always (often its percentage), but it can be in absolute terms so let's keep it!

dashboard-frontend/src/fixtures/FakeStudyBuilder.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  private getCreationDate() {
+    if (!this.creationDate) {
+      throw this.requiredCallError('setCreationDate');
+    }
+
+    return this.creationDate;
+  }
+
+  private getDesiredParticipants() {
```
nandanrao, line 85: I have the sense you've been indoctrinated into Java somehow.

We'll need to unindoctrinate you! 

All you're doing here is validating a config. Rather than create methods for all the config params (Java's "Builder" pattern), you can just create a "config" object (this is super well supported in JS, this is why ES6 went to all the crazy trouble to support defaults inside objects) or just use named params with types or whatever. If you have a config, you just run a single loop over the config and check for required fields (or just json schema or, actually, use your type system to catch missing fields at compile time, given that you have a type system, no?). 

That will remove about 120 lines to this file and get you the same result ;)

nandanrao (changes requested):

dashboard-frontend/src/types/study.ts
ericrfdev: Mmmm oki, I assumed that we had this information and I wanted to use it to display the currency symbol in the UI alongside **currentBudget** and **currentPricePerParticipant**

But you're right, thinking again about it I don't think this is needed for this kind of app. This info is more useful for apps that sell something to their users. I'll remove it 👍

dashboard-frontend/src/types/study.ts
ericrfdev: Perfect, I'll keep it then 👍

@nandanrao  Do we have the percentage always ? or it is also missing sometimes ?

If it's always there we can use it to always compute desiredParticipants at the api level, but if it's not always there I need to update the types to take into account this.

dashboard-frontend/src/fixtures/FakeStudyBuilder.ts
ericrfdev: Yes, we can get rid of it 👍

I ended up reaching for this pattern to make it more expressive and clear on the callsite, but you're right it ended up being an overkill. 😞

I'll refactor it into a function that accepts a config object and I'll rely on the type system to ensure that all the required config is provided.

ericrfdev:
@nandanrao I added an extra commit and fixed the things commented in the review:
- Remove curency from study
- Remove builder pattern

After approval I'll squash both commits before merging

dashboard-frontend/src/helpers/strings.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { createSlugFor } from './strings';
+
+describe('createSlugFor', () => {
```
nandanrao, line 3: This is cool. I like the idea of being able to use this during configuration time!

dashboard-frontend/src/helpers/numbers.ts
```diff
@@ -0,0 +1,74 @@
+export const calculatePercentage = ({
```
nandanrao, line 1: Great set of helpers, well tested! Will be useful.

dashboard-frontend/src/fixtures/study.ts
```diff
@@ -0,0 +1,213 @@
...
+  calculatePercentage,
+  round,
+} from '../helpers/numbers';
+import { createSlugFor } from '../helpers/strings';
+import { lastValue } from '../helpers/arrays';
+import { StratumProgress, Study, StudyProgress } from '../types/study';
+
+const chance = Chance();
+
+export const createFakeStudy = ({
```
nandanrao, line 13: Looks straightforward to me!

nandanrao (approved):
Looks good, we'll need to do some changes on the study types as you brought up, but this part looks great.

dashboard-frontend/src/types/study.ts
nandanrao: Ok - sorry I wasn't clear on this before: 

 `desiredParticipants?` should be optional in both the types, on study level and stratum level!

The rest we should always have. 

One option is to tell us how many they want, another is to say "just get as many as you can, given this perentage balance".

dashboard-frontend/src/helpers/numbers.ts
ericrfdev: I'm glad that you like them 🙌

dashboard-frontend/src/types/study.ts
ericrfdev: Oooh this last paragraph is key, now everything is clear 👌

I'll make both of them optional. Thanks!!

ericrfdev:
@nandanrao I already squashed the commits and I'm merging the PR. (the tests are passing and manually verified the changes too)

The only code I changed from your last review is making the **desiredParticipants** optional. (what we discussed previously in a comment)

I also had to update the code slightly to preserve the type safety. The commit with the changes is this one: https://github.com/vlab-research/vlab/pull/21/commits/92d6d1746e22f9f8d380786bedb57f7dabdc4047 in case you want to look at the diff.

Issue #24, opened 2021-07-01
Create Studies Page


ericrfdev:
### Acceptance criteria
- Given an authenticated user 
  - When he hasn't created any Study and visits the home page
    - He sees information on how to create a Study
  - When he has created one Study and visits the home page
    - He sees the name and creation date of the Study
  - When he has created 20 Studies and visits the home page
    - He sees a list of 10 studies
    - He sees the 20 studies after paginating once 
  - When he visits the home page and there is an error while fetching the Studies
    - He sees an error page with a button to retry fetching the Studies

ericrfdev:
I expect runtime errors in the render methods of the components to be strange because of Typescript plus the automatic tests. But I created the following ticket to deal with them #25 in case one happens.

PR #26, opened 2021-07-12
(#24) Create Studies Page

ericrfdev:
Finally!! 😊

@nandanrao I recommend to go commit by commit, from oldest to newest.

Each commit have a brief explanation about the changes. 

After review, before merging the PR, I'll squash them all. 😉


dashboard-frontend/cypress/integration/listing-studies.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        );
+      });
+
+      cy.visit('/');
+
+      assertLoaderAppears();
+    });
+
+    it('He sees an error page with a button to retry fetching the Studies', () => {
+      cy.contains('Something went wrong while fetching the Studies.').then(
```
nandanrao, line 90: This is great, but can we add the actual error message from the server? Even if they don't use it, it helps users to know if they're seeing the same message what to do.

nandanrao (changes requested):
Hey man, 

Sorry this took so long. I'm really happy with this! I think it's great. I'm not totally convinced by infinite scroll, sometimes I feel like it's complexity for flashiness, but let's keep it for now (it certainly is flashy!)

The tests are great, nice job setting up e2e tests w/ cypress and making them effective. They were really nice to read!

I put a small request on error message, but that's all!

nandanrao:
BTW - logo is cool! :D

dashboard-frontend/cypress/integration/listing-studies.spec.ts
ericrfdev: Yes definitely 👌  I'll display the error message from the server if available and fallback to the generic message in case none is provided or the error is a timeout.

ericrfdev:
Hey!!

I'm glad that you liked the PR and the tests 🙌

About the infinite scroll I ended up going for it because it's very common nowadays, lots of apps have feeds and people are used to them, myself included. Also on mobile I find the experience a little better compared to pagination.

But as you said we can revisit this in the future and simplify/change things. 👌

dashboard-frontend/src/pages/StudiesPage/useStudies.tsx
```diff
@@ -0,0 +1,95 @@
...
+    studies: (query.data || [])
+      .flatMap(page => page)
+      .map(studyResponse => studyResponse.data)
+      .flatMap(studyData => studyData),
+    errorMessage: query.error?.message || defaultErrorMessage,
+  };
+};
+
+/**
+ * TODO:
```
nandanrao, line 33: Great job implementing exactly what's needed for now with reasonable TODO comments for how it should be generalized/abstracted when the time comes.

nandanrao (approved):
Looks awesome!

Issue #27, opened 2021-07-14
Create Study Page


ericrfdev:
### Acceptance criteria
- Given an authenticated user 
  - When he has created one Study and visits the home page
    - He's redirected to the Study page when clicking the created Study 
  - When he visits a specific Study page
    - He sees the name of the Study
    - He sees an overview of the current Study progress
    - He sees an areachart with current participants data over time
    - He sees current avg. deviation data over time in the areachart after clicking the related stat card
    - He's redirected to the Studies page when clicking the back button
  - When he has created a Study with one Segment and visits the Study page
    - He sees a table that displays summary per segment data
    - He sees a table that displays percentage of participants per segment data
    - He sees a table that displays participants per segment data
  - When he has created a Study with 12 Segments and visits the Study page
    - He sees 7 Segments in the Summary per Segment table
    - He sees the remaining 5 Segments after clicking the Next button
  - When he visits a specific Study page and clicks the Studies link in the header
    - He's redirected to the Studies page
  - When he visits a specific Study page and there is an error while fetching the Study
    - He sees an error page with a button to retry fetching the Study

Issue #11, opened 2021-06-08
Create a low fidelity mockup of the application


ericrfdev:
The purpose of the low fidelity mockup is to iterate it as needed to answer:

- Which features do we want for the MVP ?
- Visual hierarchy for each feature or main area of the app

and to be in the same page about the expected layout of the app.

ericrfdev:
Hi @nandanrao !! 
I created a couple of wireframes for the **Studies** and **Study Page** which are the UI pages that are going to be tackled first.
I also defined some of the behaviours and interactions of the designed elements in the wireframe which will be helpful to have in development type.

**Studies page:** This page will display the name and creation date for the different studies
- Newest studies comes first
- Infinite scrolling to display all the studies (without virtual list for now)
- When loading more studies, a loading indicator is displayed at the end of the list
- When the tab is active, in the background we'll periodically poll the studies endpoint for newest studies. If new ones are found, will be added automatically to the top of the list and we'll let the user know that this happened
- In Desktop and Table Screen Sizes:
-  - The content is displayed inside a centered box which grows as more studies are displayed (instead of being fixed size with scrolling inside)
- - The whole page scrolls, the title of the page is not fixed. The only thing that is fixed is the sidebar
- - The name of the studies is truncated after two lines(I'm assuming names of studies are big, but if it's not the case @nandanrao we will truncate at 1 which facilitate things)
- In Mobile Screen Sizes:
-  - The list of studies is displayed without a box
-  - The name of the studies is truncated after X lines
-  - The creation date of the studies is listed either at the right of the study row like in desktop or below the name (pick the one that looks best when implementing it)
- No sorting functionality for now.
- - This will require some extra backend and frontend work and I think this feature is not necessary for now
- - The default order needs to be, newest studies comes first
- No search functionality for now
- - This will require some extra backend and frontend work and I think this feature is not necessary for now
- - We'll not have people with lots of studies at first to justify the feature at this stage

![image](https://user-images.githubusercontent.com/84807811/122688253-84450d80-d21b-11eb-8d08-9e7dbe92f396.png)

**Study page:** This page will display all the data related to an specific study and the different segments of participants (Stratum) it has defined.
- back button at the left of the title
- the studies page link in the sidebar is not selected anymore
- date-range selector at the top-right of the page
- title, back button and date-range selector and not fixed, whole page is scrollable except for the sidebar
- There is a centered box like in the studies page.
[...]

Issue #3, opened 2021-05-25
Visualize recruitment by strata


nandanrao:
As a user running a study, I want to be able to visualize how well recruitment is progressing, by strata, from dummy data. 

In particular, I want to know: 

1) Is the campaign progressing well, will it end well?
2) Are there individual strata that are not progressing well?

"How well" is determined by distance from a "goal" number of respondents in each strata. This "goal" for each strata is determined by the optimization routine and can change each time the routine runs.

This should work with dummy data hardcoded into the frontend, without concern for how the data gets there. 

Here is the current Grafana dashboard that serves this purpose: 

![2021-05-31-220636_grim](https://user-images.githubusercontent.com/7491490/120238730-12515800-c25d-11eb-84bf-29c10a1ef0b8.png)

![2021-05-25-153108_grim](https://user-images.githubusercontent.com/7491490/119506873-9d6ab380-bd6e-11eb-8fbf-c2ff7caae8bc.png)

![2021-05-25-153122_grim](https://user-images.githubusercontent.com/7491490/119506868-9cd21d00-bd6e-11eb-8f2f-7a9288830c49.png)

nandanrao:
Current average deviation: 

```SQL
WITH tt AS
  (WITH t AS
    (SELECT row_to_json(json_each(details)) d, created
     FROM (SELECT details, adopt_reports.created
          FROM adopt_reports
          JOIN campaigns ON campaignid = campaigns.id
          WHERE campaigns.NAME = '$campaign'
          ORDER BY created DESC
          LIMIT 1))
  SELECT
      created,
      (d->'value'->>'goal')::FLOAT AS goal,
      (d->'value'->>'respondent_share')::FLOAT AS share
  FROM t)
SELECT AVG(ABS(goal - share))*100 AS dif
FROM tt;
```

nandanrao:
![2021-05-31-230755_grim](https://user-images.githubusercontent.com/7491490/120242402-16817380-c265-11eb-8dfb-16eed950b616.png)

PR #28, opened 2021-08-15
(#27) Create Study Page

ericrfdev:
@nandanrao Study Page ready 🥳

A bit of a summary about what to expect:
- The stats card at the top of the page are displaying the current Study progress
- The chart is displaying the progress of the Study since it's creation date
- The tables are displaying the current progress for each of the segments of an Study.
- All the previous visual elements are polling the api every 30 seconds for updates. (only 2 request since the stats and the chart are sharing the same query).
- I didn't add the possibility to view data for a specific date range in this PR. After this one is merged, we can talk about it. I already have an idea on how to approach it, but wanted to clarify some doubts with you first. 👍

### DIFF
--- a/dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
+++ b/dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
@@ -1,0 +1,425 @@
+import { Response } from 'miragejs';
+import Chance from 'chance';
+import { createFakeStudy } from '../../src/fixtures/study';
+import {
+  makeServer,
+  createStudyResource,
+  createStudyProgressResource,
+  createSegmentProgressResource,
+} from '../../src/server';
+import { StudyResource } from '../../src/types/study';
+
+const chance = Chance();
+
+describe('Given an authenticated user', () => {
+  let server: ReturnType<typeof makeServer>;
+
+  beforeEach(() => {
+    server = makeServer({ environment: 'test' });
+  });
+
+  afterEach(() => {
+    server.shutdown();
+  });
+
+  describe('When he visits a specific Study page', () => {
+    beforeEach(() => {
+      const study = createStudyResource(server, {
+        name: 'Weekly consume of meat',
+        slug: 'weekly-consume-of-meat',
+      });
+
+      [
+        {
+          id: chance.guid({ version: 4 }),
+          datetime: new Date('2021-06-19').getTime(),
+          currentParticipants: 0,
+          expectedParticipants: 0,
+          currentAverageDeviation: 0,
+          expectedAverageDeviation: 0,
+          desiredParticipants: null,
+        },
+        {
+          id: chance.guid({ version: 4 }),
+          datetime: new Date('2021-06-20').getTime(),
+          currentParticipants: 15037,
+          expectedParticipants: 16000,
+          currentAverageDeviation: 1,
+          expectedAverageDeviation: 1.1,
+          desiredParticipants: null,
+        },
+        {
+          id: chance.guid({ version: 4 }),
+          datetime: new Date('2021-06-21').getTime(),
+          currentParticipants: 20137,
+          expectedParticipants: 21250,
+          currentAverageDeviation: 1.5,
+          expectedAverageDeviation: 1.7,
+          desiredParticipants: null,
+        },
+      ].forEach(progress => {
+        createStudyProgressResource(server, {
+          study,
+          progress,
+        });
+      });
+
+      cy.visit('/studies/weekly-consume-of-meat');
+
+      assertLoadersAppear();
+    });
+
+    it('He sees the name of the Study', () => {
+      cy.contains('Weekly consume of meat');
+    });
+
+    it('He sees an overview of the current Study progress', () => {
+      cy.contains('Current Participants');
+      cy.contains('20,137');
+
+      cy.contains('Expected Participants');
+      cy.contains('21,250');
+
+      cy.contains('Current Avg. Deviation');
+      cy.contains('1.5 %');
+
+      cy.contains('Expected Avg. Deviation');
+      cy.contains('1.7 %');
+    });
+
+    it('He sees an areachart with current participants data over time', () => {
+      assertStudyProgressChartAppears({
+        numDataPoints: 3,
+      });
+    });
+
+    it('He sees current avg. deviation data over time in the areachart after clicking the related stat card', () => {
+      assertStudyProgressChartAppears({
+        numDataPoints: 3,
+      });
+      cy.get('[data-testid="study-progress-chart"]').contains('20,000');
+
+      cy.get('[data-testid="study-page-stats"]')
+        .contains('Current Avg. Deviation')
+        .click();
+
+      assertStudyProgressChartAppears({
+        numDataPoints: 3,
+      });
+      cy.get('[data-testid="study-progress-chart"]')
+        .contains('20,000')
+        .should('not.exist');
+      cy.get('[data-testid="study-progress-chart"]').contains('1.5');
+    });
+
+    it("He's redirected to the Studies page when clicking the back button", () => {
+      cy.get('[data-testid="back-button"]').click();
+
+      cy.url().should('eq', `${Cypress.config().baseUrl}/`);
+    });
+  });
+
+  describe('When he has created a Study with one Segment and visits the Study page', () => {
+    beforeEach(() => {
+      const study = createStudyWithRandomProgress(server);
+
+      createSegmentProgressResource(server, {
+        study,
+        segmentProgress: {
+          id: 'e1912f80-d9ff-4680-8bbd-bdefd1834b38',
+          name: '64-spain',
+          currentBudget: 7200,
+          desiredPercentage: 10,
+          currentPercentage: 5.95,
+          expectedPercentage: 6.55,
+          desiredParticipants: null,
+          currentParticipants: 1429,
+          expectedParticipants: 1571,
+          datetime: Date.now(),
+          currentPricePerParticipant: 2,
+          percentageDeviationFromGoal: 5.95,
+        },
+      });
+
+      cy.visit(`/studies/${study.slug}`);
+
+      assertLoadersAppear();
+    });
+
+    it('He sees a table that displays summary per segment data', () => {
+      cy.contains('Summary per segment');
+
+      cy.get('[data-testid="summary-per-segment-table"]').contains('Name');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('%Progress');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('Budget');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('Spent');
+
+      cy.get('[data-testid="summary-per-segment-table"]').contains('64-spain');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('94.05');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('7,200');
+      cy.get('[data-testid="summary-per-segment-table"]').contains('2,858');
+    });
+
+    it('He sees a table that displays percentage of participants per segment data', () => {
+      cy.contains('% Participants per segment');
+
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('Name');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('%Desired');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('%Current');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('%Expected');
+
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('64-spain');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('10');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('5.95');
+      cy.get(
+        '[data-testid="percentage-participants-per-segment-table"]'
+      ).contains('6.55');
+    });
+
+    it('He sees a table that displays participants per segment data', () => {
+      cy.contains('Participants per segment');
+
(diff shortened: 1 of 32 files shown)


dashboard-frontend/src/server.ts
```diff
@@ -80,6 +148,101 @@ export const makeServer = ({ environment = 'development' } = {}) => {
...
+        const study = (db.studies as any).findBy({ slug: request.params.slug });
+
+        return {
+          data: study,
+        };
+      });
+
+      /**
+       * Things to consider when implementing the real endpoint:
+       *  - When a Study is initially created, an associated StudyProgress must be created with default values.
```
nandanrao, line 162: I don't like this as a requirement. A study that has been created but doesn't yet have any progress seems like a valid database state and I think enforcing that to be invalid is brittle. It might be useful to create an initial "studyprogress" with the progress at time 0 when a study has been created, but that shouldn't be a _requirement_ for the frontend. 

The frontend should therefore be able to handle the state where there is no studyprogress. That could mean that the whole page is replaced by text saying "Nothing to report yet! Check back soon", which would be fine.

dashboard-frontend/src/server.ts
```diff
@@ -80,6 +148,101 @@ export const makeServer = ({ environment = 'development' } = {}) => {
...
+       *    At the moment, we only display all-time progress (chart) or current progress (stat cards).
+       *      - So this endpoint should only return one data point per day (latest one for that day)
+       *      - This endpoint is not paginated because we need to display all available data in the chart.
+       *        If we anticipate that this will be a problem, we can limit the number of StudyProgress returned
+       *        by this endpoint to a fixed number. (first study progress and latest one will always be included)
+       *
+       *  - Soon, when we allow the users to pick a specific range of dates, we'll add the required query_parameters.
+       *    To decide yet if we want more granularity and allow the user to see hourly and minutes updates reflected in the chart.
+       */
+      this.get('/studies/:slug/progress', ({ db }, request) => {
```
nandanrao, line 176: `StudyProgressResource` and `StudySegmentProgressResource` are representations that are useful for the dashboard and progress tracking. But they are not good representations for the rest of the system. In particular, they're completely redundant, `StudyProgressResource` is entirely calculable from `[]StudySegmentProgressResource`! It's just a view on that data that is of interest to the user. It's not data in and of itself and certainly shouldn't be fetched separately. 

So definitely `/progress` shouldn't need to exist.

We should have a translation layer in the frontend that translates the endpoint/backend representation into "StudyProgressResource" and "StudySegmentProgressResource" which make sense for dashboarding. 

I think the backend representation probably looks something like this: 

```javascript
interface ??? {
  id: string;
  name: string;
  datetime: number;
  current_budget: number;
  desired_percentage: number;
  expected_percentage: number;
  current_percentage: number;
  desired_participants: number | null;
  expected_participants: number;
  current_participants: number;
  current_price: number;
}
```

Where `percentageDeviationFromGoal` is a front-end only concept which is easily calculable. We can also add `deviationFromGoal` in terms of absolute participant numbers as well. And this provides a place of abstraction to add more ideas that matter for visualization but that aren't data per se. 

And actually, I suspect we will be adding more to the backend representation, like `amount_spent`, etc., but the frontend shouldn't need to worry about using it quite yet.

dashboard-frontend/src/pages/StudyPage/Table.tsx
```diff
@@ -0,0 +1,190 @@
...
+            {children}
+          </table>
+          <Pagination {...pagination} />
+        </div>
+      </div>
+    </div>
+  </div>
+);
+
+const Pagination = ({
```
nandanrao, line 121: I'm suprised you had to build a table component from scratch. Tailwind doesn't give you one?

In any case, they look and work well, so I'm happy with them! And I like being able to customize their styling more in the future because they're custom. 

That being said, we'd definitely like to add the ability to sort by column at some point. Maybe also change the number viewed per page. It's all pretty traditional stuff, and shouldn't be hard to implement, but worth keeping in mind.

dashboard-frontend/src/pages/StudyPage/StudySegmentsTables.tsx
```diff
@@ -0,0 +1,72 @@
...
+  showLoader: boolean;
+  studySlug: string;
+}) => {
+  const sharedProps = {
+    ...props,
+    loaderTestId: 'study-segment-table-skeleton',
+  };
+
+  return (
+    <React.Fragment>
```
nandanrao, line 15: Looking at this now, I think it's probably more useful to have these in some sort of "tab" type control, so that you look at one or the other, but you see more rows by default. I think this is perfect for now, but can you make a new issue to throw in the backlog for something like this?

dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
```diff
@@ -0,0 +1,425 @@
...
+            }
+          )
+      );
+
+      cy.visit('/studies/weekly-consume-of-meat');
+
+      assertLoadersAppear();
+    });
+
+    it('He sees an error page with a button to retry fetching the Study', () => {
```
nandanrao, line 310: Great.

dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
```diff
@@ -0,0 +1,425 @@
...
+      cy.contains('Current Participants');
+      cy.contains('20,137');
+
+      cy.contains('Expected Participants');
+      cy.contains('21,250');
+
+      cy.contains('Current Avg. Deviation');
+      cy.contains('1.5 %');
+
+      cy.contains('Expected Avg. Deviation');
```
nandanrao, line 86: Testing the number formatting is great. Testing the exact wording of subtitles feels a bit brittle to me. It's not excessive here, but something to keep in mind.

nandanrao (approved):
I think this is brilliant. Really nice. 

I think my biggest concern in just to define the backend/api. I think it's clear that it's time to actually define/build the database and server :/.

dashboard-frontend/src/pages/StudyPage/StudySegmentsTables.tsx
ericrfdev: Yes I think it's a good idea to think in ways to improve how we present the current progress of the different Segments of an Study. 👌

 I created an issue for it: https://github.com/vlab-research/vlab/issues/29

I already provided some details in the issue, but to give you some context, I went for 3 tables of 4 columns because this way the information can be consumed fine in any screen size. I tried with more columns but the table broke on mobile screen sizes 😢

dashboard-frontend/src/pages/StudyPage/Table.tsx
ericrfdev: I'm glad that you liked them 🙌

Answering your question, TailwindUI provides me with the raw separated JSX for the table and the pagination, the rest is up to me.

I wanted to have simpler version ready before going for the sorting, but we can definitely add it up soon. 👍

dashboard-frontend/src/server.ts
ericrfdev: I understand your point that these representations will only be used by the dashboard and due this reason, at first sight it seems that make more sense to receive the raw data from the backend and do the computation on the frontend.

Let's think what will happen, if we follow this approach and we try to represent the current UI:
- Good
  - We're still able to represent the tables

- Normal/Bad
  - We're still able to represent an overview of the current progress of an Study (stats cards ui). 
    But in order to do this, we need the current progress for all the segments of an Study, this translates into one of the following scenarios:
    - No pagination is required since we need all the data to compute the overview. All the data is refetched every X seconds leading into some overfetching.
    - We still use pagination to avoid refetching all the data every X seconds (we only fetch the latest data)
      - this complicates the frontend code a lot by introducing lots of conditional logic that involve coordination of requests

- Bad
  - The chart cannot be represented anymore. The reason why is:
    - The chart is representing data points for a specific piece of data, at a specific time. This piece of data is computed from the progress of the all the Segments of an Study at that same specific time.
      - To represent all-time data, it means fetching all the segments progress since the start from an Study

Above I listed some of the problems that I anticipate, but not all of them (due to time constraints).

I agree with you that normally it's not a good idea to reach for extra representations at the **api level** only to fulfill the needs of a single client, but I think this rule should not be set on stone and when justified such as in this case, we can make an exception.

dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
ericrfdev: To give you some context, I wrote this test prior to the implementation, and used all these assertions to drive the implementation. That's the main reason why are there.

From the regression test point of view I agree with you that the assertions about the number formatting are more valuable because they're testing more things that can go wrong (existence + formatting logic). But for me the whole functionality are both the values and the labels, if a label is not rendered anymore I also want to be warned about it, because a number without a label does not tell the user anything.

Actually this test is in my list of possible improvements, what I think it can be improved about it is:
- Testing for the visibility of labels/values instead of only testing that are rendered
- Test that the different values are matched with the right label in a specific stat card

I haven't done these improvements because there were many things more important to do, and apart from time, some of them comes with extra tradeoffs such as adding extra code to facilitate tests (aka testids).

At the end it all comes down to how you feel about the tests, the usefulness and confidence that provides to you. It can always be better and may change overtime.

dashboard-frontend/src/server.ts
ericrfdev: I agree, good point 👌

I created an issue to handle this: https://github.com/vlab-research/vlab/issues/31

nandanrao: Cool, thanks!

dashboard-frontend/src/pages/StudyPage/Table.tsx
nandanrao: Cool cool, thanks!

dashboard-frontend/src/pages/StudyPage/StudySegmentsTables.tsx
nandanrao: Couple thoughts on this: 

1. I think we need to support tables with more than 4 columns. I don't see a way around that. 

2. Really, nobody should ever need to look at the details of a table on their mobile. If they're looking on their mobile, it's to get an overview. If they're want to dive into the details of a table, it's because they want to see all the info of a table and this is never going to be pleasant on a phone. 

3. Let's make it not "break", but that doesn't mean it has to be an amazing experience. I have the Google Sheets app on my phone. It sucks looking at a Sheet on my phone. I have to scroll left <> right to see the columns. But it works, it's not broken, it's just inherently not an amazing experience. 

4. Options to solve the problem: some way to select which columns are shown. On a mobile, you should select fewer columns.

But really, I don't want to prioritize viewing a table on a mobile. That seems like a rare edge case to me and this can always be determined empirically when we have usage data.

dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
nandanrao: I agree with all this and definitely agree with test-first development. Some tests are a bit brittle, or do need to be changed often, and that's OK. But it's worth keeping in mind that probably I don't care what the exact label name is, I just care that the label appeared. 

Simple solutions to this involve just using a variable - putting all microcopy in a single module or config file and using tests to test that what is configured, appeared, without the app caring about what is configured.

dashboard-frontend/src/server.ts
nandanrao: These are valid points, but I think I'm only convinced by them if we decide that this page doesn't need ALL the data from ALL the segments over ALL time in order to render things for the user. 

That would be bad as a requirement because It's a lot of data. But this is a data dashboard, so maybe that's ok. I doubt the transfer of the payload from the server to the client will ever be the largest bottleneck, as opposed to server calculating things, etc. 

It would be good as a requirement because it allows us to be really flexible and powerful about what we show on this dashboard, without the backend/interaction changing. We can show anything, because we have all the data, so we can calculate anything. It's all within the power of the dashboard code, because the dashboard code is the only one that cares about presentation/visualization of data.

One big question: what did we decide about the graph in the 2nd screenshot here, I can't remember: https://github.com/vlab-research/vlab/issues/3

That graph plots ALL the data. It's also the graph I use to show people how things work a lot, I think it really helps people understand what is happening, so I value it.

dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
ericrfdev: Ah oki, I thought you were referring that having these assertions was not useful.

Now I understand you perfectly, and yes I agree that it would be an improvement to have a shared variable between the static microcopy in test assertions and the real implementation.

I'll keep this in mind for further developments and I will add a config file to store the microcopy as you suggested.

Thanks 🙌

dashboard-frontend/src/pages/StudyPage/StudySegmentsTables.tsx
ericrfdev: Okay makes sense👌  let's go with this approach of having a single table, if it does not fit in the screen, we allow the user to scroll it horizontally as you pointed.

You're right that users won't probably use the app through mobile, and it's better to optimize for the desktop experience for now.

dashboard-frontend/src/server.ts
ericrfdev: Yes it's true that by returning raw data from the api and performing the computations on the dashboard, we'll need to touch the api code less often when new requirements arrive or existent one change for the dashboard which is good.

I have been thinking more about my main concern that was the over-fetching of data and maybe if that becomes a problem, we can limit the results returned to a fixed number. For example given an specific date range, the first and latest result for that range are always returned, and through some logic we pick the rest of results to return.

So yes, I think the single endpoint can work but it will be useful to discuss it thoroughly to not shoot ourselves in the foot too early. Let's talk about this in a call. 

About the second graph, before I started with the design of the app (https://github.com/vlab-research/vlab/issues/11) we decided that it will come after the other things (stat cards, area chart, tables).

But now that the other things are bootstrapped, after defining the desired outcome with you, I can start implementing it. I'll ping you in slack to talk about this.

ericrfdev:
Last commit before squashing -> 742716522ce86833f6270d90c8c4cc9bddf7cb8a

Issue #36, opened 2021-09-03
Setup authentication and authorization in the dashboard


ericrfdev:
Setup authentication and authorization in the dashboard using auth0.

## Previous related tickets
[Spike to choose a solution/product for managing authentication and authorization](https://github.com/vlab-research/vlab/issues/12)

ericrfdev:
@nandanrao I created 3 new tickets to deal with bugs, limitations and general improvements related to the integration with Auth0.
- [Update Auth0 integration to work properly in Safari](https://github.com/vlab-research/vlab/issues/41) (bug)
- [Configure external email provider in Auth0](https://github.com/vlab-research/vlab/issues/40) (limitations)
- [Rebrand Auth0 provided pages and emails](https://github.com/vlab-research/vlab/issues/39) (general improvements)

PR #42, opened 2021-09-05
(#36) Setup authentication and authorization in the dashboard using Auth0

ericrfdev:
(#36) Setup authentication and authorization in the dashboard using Auth0

  - Install Auth0 react sdk and setup environment variables for both dev and prod environments
  - Modify server.ts
  - to perform authenticated request checks in the different endpoints
  - to stop proxying any request for the Auth0 domain
  - Show the same loader when the react app is being bostraped (index.html) and while the auth0 sdk finish checking if there is an authenticated user (App.tsx)
  - Add a png version of the logo. Auth0 universal login page will fetch it from the github cdn
  - Setup route guards for all pages (authenticated ones and login one)
  - Display the name/email and avatar of the authenticated user in the header. Clicking the avatar opens a dropdown with a sign out button that starts the logout flow
  - Create login page to start the login flow
  - Update api helpers and queries to pass the access token to the authenticated endpoints
  - Use headless chrome for e2e, disable video recording and remove one flacky redundant assertion
  - Create useAuthenticatedApi hook to use it for all interactions with authenticated endpoints
  - Create useAuth0 hook to wrap the original one, so we can provide hardcoded values during e2e (taking into account pros and cons, I prefer this over the alternatives which among other things, involves either requests to auth0 api or real interactions with the universal page)


dashboard-frontend/public/index.html
```diff
@@ -26,10 +26,108 @@
     -->
     <title>React App</title>
     <link rel="stylesheet" href="https://rsms.me/inter/inter.css" />
+    <style>
```
nandanrao, line 29: It's fine that this is in the base html when the site is built, but can't it be included as a fragment somehow? That way we can reuse it? Not urgen. 

But it's very cool looking.

dashboard-frontend/src/hooks/useAuth0.tsx
```diff
@@ -0,0 +1,33 @@
...
+  | 'isAuthenticated'
+  | 'isLoading'
+  | 'user'
+  | 'logout'
+  | 'getAccessTokenSilently'
+  | 'loginWithRedirect'
+> => {
+  const auth0Context = useRealAuth0();
+
+  if (process.env.REACT_APP_RUNNING_IN_E2E_MODE) {
```
nandanrao, line 14: Really don't love this. It's a mock. But instead of the testing program being in charge of creating the mock or the app having some sort of inversion of control that makes it easy to mock, the app code itself checks for an env variable and then returns a mock! 

:/

dashboard-frontend/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <div className="bg-white shadow overflow-hidden rounded-none sm:rounded-md pt-8 pb-10">
+          <div className="flex flex-col items-center">
+            <WelcomeIllustration className="h-56 w-72" />
+
+            <p className="pt-10 pl-9 pr-9 sm:pl-20 sm:pr-20 text-4xl text-gray-900 text-center">
+              Ready to start?
+            </p>
+
+            <p className="pt-5 pl-9 pr-9 sm:pl-20 sm:pr-20 text-xl text-gray-900 text-center">
+              Log in now to view the progress of your Studies!
```
nandanrao, line 25: Will have to work on this copy at some point but it's good for now ;) (they will also be able to launch studies, not just view the progress!!)

nandanrao (approved):
Overall looks great! I made one comment about the mocking of the auth response. Feel free to open an issue and we can prioritize it some other time.

dashboard-frontend/src/pages/LoginPage/LoginPage.tsx
ericrfdev: Uups yes you're right, I'll add this to my todo list.

Thanks 👍

dashboard-frontend/public/index.html
ericrfdev: I just read the create-react-app docs (https://create-react-app.dev/) and didn't saw that they offer this out of the box.

It can be archived by taking advantage of npm scripts or by ejecting from create-react-app and using a Webpack plugin.

Since these styles and the html markup are only to replicate the LoaderPage, I don't think it's a big problem maintain these two in sync, since it's not something that changes often.

A cheap improvement that I can do at some point is to update the LoaderPage to use the style declarations defined here, making more explicit the relation between the two. I can also write a section about this in the README.md when I create it.

dashboard-frontend/src/hooks/useAuth0.tsx
ericrfdev: Yes I agree with you that this is not ideal.

The thing is that the app runs on a separate process than Cypress, so out of the box is not possible to mock a module.

This mean that in case you want to change the default implementation of something during Cypress tests, you'll need to instrument the code always.

Right now the fake implementation that is used when Cypress tests are running is hardcoded, but this control can be given to Cypress if needed. Instead of returning this hardcoded object, the app code during Cypress tests could read a property from the Windows object which Cypress is able to modify to provide the right mock.

At the moment I don't mind having it like this, since I want to have this auth state for all the e2e tests, in the future when we update some of the e2e tests to talk to the real backend, we can think about improving this.

Before going with this approach I did some research and the Cypress team is already working on an interesting solution for this. It's still in alpha but I'm watching they git repo for updates. Take a look at this when you can:
- https://www.youtube.com/watch?v=RAFdYqRO2vI
	- https://github.com/bahmutov/mock-in-bundle

PR #45, opened 2021-09-20
(#44) Refactor code to allow the StudyPage to obtain most if the info…

ericrfdev:
Refactor code to allow the StudyPage to obtain most if the info it requires from the segments-progress endpoint

  - Modify "/studies/:slug/segments-progress" mirage endpoint to return all the segments progress since the start of the Study.
  - Update Study Page to rely on a single useStudy hook that provides all the data required by the page.
    - Most of the data is computed from the segments-progress endpoint
    - The Study endpoint is also used to retrieve the name of the Study.
  - Delete dead code
    - hooks
    - mirage endpoints
    - api helpers
    - api types
  - Handle empty states for the different elements visible in the Study page
  - Update createFakeStudy function to work in days instead of hours


dashboard-frontend/cypress/integration/viewing-a-study.spec.ts
```diff
@@ -321,6 +358,43 @@ describe('Given an authenticated user', () => {
       cy.contains('Weekly consume of meat');
     });
   });
+
+  describe("When he visits a specific Study page that hasn't any progress", () => {
```
nandanrao, line 362: Looks good!

dashboard-frontend/src/helpers/study.ts
```diff
@@ -0,0 +1,25 @@
...
+    0
+  ),
+  currentAverageDeviation: calculateAverageDeviation(
+    stratumProgressList.map(
+      ({ percentageDeviationFromGoal }) => percentageDeviationFromGoal
+    )
+  ),
+  expectedAverageDeviation: calculateAverageDeviation(
+    stratumProgressList.map(({ expectedPercentage }) =>
+      round(100 - expectedPercentage)
```
nandanrao, line 22: Why is this 100 - expectedPercentage? That would imply that 100 is the desiredPercentage, no? But that's not the case.

dashboard-frontend/src/pages/StudyPage/useStudy.tsx
```diff
@@ -0,0 +1,96 @@
...
+    () => computeStudyProgress(currentSegmentsProgress),
+    [currentSegmentsProgress]
+  );
+
+  const isLoading = !studyQuery.data || !studySegmentsProgressQuery.data?.data;
+
+  const anyErrorDuringLoading =
+    isLoading && (studyQuery.isError || studySegmentsProgressQuery.isError);
+
+  return {
```
nandanrao, line 43: I can't see this type anywhere, but it seems like a pretty important thing to have as a type, as you're using it everywhere, no? this "study" object?

nandanrao (commented):
Overall, looks pretty good! Left a couple of questions

dashboard-frontend/src/helpers/study.ts
ericrfdev: Yes it's a mistake, good catch 👍

dashboard-frontend/src/pages/StudyPage/useStudy.tsx
ericrfdev: The type is inferred, no need to type it explicitly 🎉  (Typescript is a structural type system).

ericrfdev:
@nandanrao Just answered your comments and updated the PR with an extra commit that fixes:

- Fix expectedAverageDeviation calculation, the result of the calculation remains the same
          but the way I did it was conceptually wrong. (**the thing you noticed**)
 - Update createFakeStudy
        - Rename 'hours' variable to 'days' to reflect new reality better
        - Fix a bug when updating the segment data. I was using the 'desiredParticipants' variable
            instead of 'desiredParticipantsPerStrata' which translated in
            incorrect numbers (although consistent between them) for:
              currentPercentage, expectedPercentage and percentiageDeviationFromGoal

After approval in case everything is ok, I'll squash the two commits before merging.

nandanrao:
Hey - so I think all the deviation things are not right -> https://github.com/vlab-research/vlab/blob/main/dashboard-frontend/src/helpers/numbers.ts#L23 this isn't a statistic we are actually interested in. 

By `deviation` in this context all we want is the difference from our desired number. So if my desiredPercentage is 2% and my currentPercentage is 3% the devation is 1%. 

So the average is the average of those deviations across segments, the expected average is the average absolute difference between the expectedPercentage and the desiredPercentage.

0x6165726F78:
@nandanrao Oks makes sense, thanks for the example 👍

I hope I get it right this time. I added an additional commit with the changes  -> https://github.com/vlab-research/vlab/pull/45/commits/a41e04f6c3867fb26fc98c1cbe38579f2d99487f

ericrfdev:
[@nandanrao](https://github.com/nandanrao) Oks makes sense, thanks for the example 👍

I hope I get it right this time. I added an additional commit with the changes -> [a41e04f](https://github.com/vlab-research/vlab/commit/a41e04f6c3867fb26fc98c1cbe38579f2d99487f)

dashboard-frontend/src/helpers/study.ts
```diff
@@ -0,0 +1,35 @@
+import { StudySegmentProgressResource } from '../types/study';
+import { calculateAverage } from './numbers';
+
+export const computeStudyProgressDataFrom = (
+  segmentProgressList: StudySegmentProgressResource[]
+) => ({
+  currentParticipants: computeCurrentParticipantsFrom(segmentProgressList),
+  expectedParticipants: computeExpectedParticipantsFrom(segmentProgressList),
+  currentAverageDeviation: calculateAverage(
```
nandanrao, line 9: Perfect!

dashboard-frontend/src/helpers/study.ts
```diff
@@ -0,0 +1,35 @@
...
+    segmentProgressList.map(({ currentPercentage, desiredPercentage }) =>
+      Math.abs(desiredPercentage - currentPercentage)
+    )
+  ),
+  expectedAverageDeviation: calculateAverage(
```
nandanrao, line 14: perfect.

nandanrao (changes requested):

dashboard-frontend/src/helpers/study.ts
nandanrao: (that's the "current_gap" from https://github.com/vlab-research/vlab/issues/3)

nandanrao:
LGTM!

PR #55, opened 2021-10-24
(#50) Bootstrap api for the dashboard by adding "study" and "health" endpoints

ericrfdev:


dashboard-api/internal/study.go
```diff
@@ -0,0 +1,49 @@
...
+func NewStudy(id, name, slug string, createdAt time.Time) Study {
+	return Study{
+		id:        id,
+		name:      name,
+		slug:      slug,
+		createdAt: createdAt,
+	}
+}
+
+func (s Study) ID() string {
```
nandanrao, line 33: Why have a getter for this, why isn't it just public? Do you really think anyone's going to overwrite it? That's a very Java-esque paradigm, I don't see the advantage.

dashboard-api/cmd/api/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,39 @@
+package bootstrap
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/kelseyhightower/envconfig"
+	_ "github.com/lib/pq"
```
nandanrao, line 9: Let's use pgx, which is what we're using everywhere else, unless there's an interesting reason to have a different driver?

dashboard-api/internal/platform/server/handler/studies/read.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/gin-gonic/gin"
+	studiesmanager "github.com/vlab-research/vlab/dashboard-api/internal"
+)
+
+type readRequest struct {
+	Slug string `uri:"slug" binding:"required"`
+}
+
+type readResponse struct {
```
nandanrao, line 15: Wait, this is the THIRD time I'm seeing the same type defined. I don't see how this is a good idea. Use an alias if you want abstraction, don't manually redefine it over and over!

nandanrao (changes requested):

dashboard-api/internal/study.go
ericrfdev: You guessed it right, I used the getters to achieve immutability.

But I agree with you, looking at it again, it's overkill and the problem they may solve, is non existent for now.

I'll update the code to get rid of them. 👍

dashboard-api/cmd/api/bootstrap/bootstrap.go
ericrfdev: I was not aware that we were using pgx. The main reason I used pq it's because it supported database/sql but it seems like pgx also supports it, so I'm happy to change the dependency. 👌

Relying on the database/sql package from the go standard library for the repositories code it's good because it provides a generic interface for any SQL database, so updating repositories implementations becomes much easier.

Pgx can be used as standalone or in combination with database/sql, I prefer the second approach for what I just mentioned above. If at some point (or you already anticipate) we want some of the additional features that the standalone pgx version offers, we can migrate the repositories implementation.

dashboard-api/internal/platform/server/handler/studies/read.go
ericrfdev: Thanks for the feedback, didn't know that go supported type aliases. 💯

Thinking more about this, for now it makes more sense to just return the entity. (and not using a type alias or extra type).

If in the future for some controller we need to return a stripped-down version of an entity, that would be a good case for an extra type (eg. User, PublicUser).

ericrfdev:
@nandanrao Updated the PR with all the suggested changes 🙌  Let me know what you think

nandanrao:
Looks simpler! Thanks!

PR #57, opened 2021-10-31
(#56) Add "studies" endpoint to the dashboard-api

ericrfdev:


dashboard-api/internal/platform/server/handler/studies/list.go
```diff
@@ -0,0 +1,113 @@
...
+			return
+		}
+
+		studies, err := studyRepository.GetStudies(ctx, pagination.Cursor, pagination.Number)
+		if err != nil {
+			ctx.Status(http.StatusInternalServerError)
+			return
+		}
+
+		data := make([]studyReponse, 0, len(studies))
```
nandanrao, line 34: Not actually a deal because the performance difference is negligible, but I find it confusing to specify the capacity and then use append. Why not just specify the length and assign?

dashboard-api/internal/platform/server/handler/studies/list.go
```diff
@@ -0,0 +1,113 @@
...
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+	studiesmanager "github.com/vlab-research/vlab/dashboard-api/internal"
+)
+
+type studyReponse struct {
```
nandanrao, line 13: You have a Study and a studyResponse that are the exact same. If you want an abstraction, why not just use a type alias and then if they ever do differ, someone can change it.

dashboard-api/internal/platform/server/handler/studies/list.go
```diff
@@ -0,0 +1,113 @@
...
+		for _, study := range studies {
+			data = append(data, studyReponse{
+				Id:        study.ID(),
+				Name:      study.Name(),
+				Slug:      study.Slug(),
+				CreatedAt: study.CreatedAt(),
+			})
+		}
+
+		if len(studies) < pagination.Number {
```
nandanrao, line 44:
```Go
var nextCursor string
if len(studies) < pagination.Number {
   nextCursor = nil
} else {
   nextCursor = createNextCursor
}
```

dashboard-api/internal/platform/server/handler/studies/list.go
```diff
@@ -0,0 +1,113 @@
...
+		return 0, fmt.Errorf("invalid cursor: %s", cursorInBase64url)
+	}
+
+	return cursor, nil
+}
+
+func createNextCursor(pagination pagination) string {
+	nextCursorAsString := strconv.Itoa(pagination.Cursor + pagination.Number)
+
+	return base64.URLEncoding.EncodeToString([]byte(nextCursorAsString))
```
nandanrao, line 112: What's the point of base64 encoding an integer?

dashboard-api/internal/platform/server/handler/studies/list_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	t.Run("should return a 500 when the repository returns an error", func(t *testing.T) {
+		studyRepository := new(storagemocks.StudyRepository)
+		studyRepository.On("GetStudies", mock.Anything, mock.Anything, mock.Anything).Return([]studiesmanager.Study{}, errors.New("unexpected-error"))
+		res := performRequest("/studies", studyRepository)
+
+		assert.Equal(t, http.StatusInternalServerError, res.StatusCode)
+	})
+
+	t.Run("should return a valid nextCursor when number of studies returned for the current page is equal to the requested ones", func(t *testing.T) {
```
nandanrao, line 59: Nice documentation explaining expected behavior at the edge case!

dashboard-api/internal/platform/storage/cockroachdb/study_repository_test.go
```diff
@@ -76,3 +76,74 @@ func Test_StudyRepository_GetStudyBySlug_Succeed(t *testing.T) {
...
+
+func Test_StudyRepository_GetStudies_UnexpectedError_during_Query(t *testing.T) {
+	offset := 0
+	limit := 20
+
+	db, sqlMock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+	require.NoError(t, err)
+
+	sqlMock.ExpectQuery(
+		"SELECT id, name, slug, created FROM studies ORDER BY created DESC OFFSET $1 LIMIT $2").
```
nandanrao, line 88: Don't repeat business logic of the unit under test in the test itself! A template and the correct args to fill it make up business logic in this case.

nandanrao (changes requested):
Overall, looks good, a few stylistic changes just to simplify and lighten up the codebase a bit.

dashboard-api/internal/platform/server/handler/studies/list.go
ericrfdev: I got rid of this code, since I'm returning the entity directly it's not necessary anymore.

dashboard-api/internal/platform/server/handler/studies/list.go
ericrfdev: For now I'll just return the entities, no extra type or type alias. Just like in the other PR -> https://github.com/vlab-research/vlab/pull/55#discussion_r743382106

dashboard-api/internal/platform/server/handler/studies/list.go
ericrfdev: In general it's good practice for the pagination cursor to be an opaque string.

This is specially helpful for public api's that have many third-party consumers.

By providing an opaque string that should not be parsed by the api consumers, the owners of the public api are able to change the underlying implementation of the pagination with more chance of avoiding unexpected breaking changes in their consumers applications.

Having said that, this api will only have a consumer for now and it's a first party application, so I'm ok if you prefer to just pass the integer and avoid this complexity for now.

Let me know what you prefer.

dashboard-api/internal/platform/server/handler/studies/list_test.go
ericrfdev: Glad that you liked it!!

dashboard-api/internal/platform/storage/cockroachdb/study_repository_test.go
ericrfdev: I'm aware of this but the mocking library forces you to write it like this.

The following is not possible:
```
sqlMock.ExpectQuery(
		"SELECT id, name, slug, created FROM studies ORDER BY created DESC OFFSET 0 LIMIT 20").
		WillReturnError(errors.New("unexpected-error"))
```

dashboard-api/internal/platform/server/handler/studies/list.go
ericrfdev: After finding that GO do not support ternary operator, the code that you provided was my first attempt but it didn't work out, because you cannot assign a **nil** to a **string variable.** That's the reason why I did it like this. (at the end, the repeated object is not that big as to impact readability, and extracting a function to returning the response and seems unnecessary)

A way to store a "value of a specific type" or "nil" in GO, is to use pointers. But for this particular scenario it will not work, because dereferencing a nil pointer will crash the application, so an extra conditional is required to check if the variable is a nil pointer. In case it's a nil pointer return it directly, otherwise dereference it.

Investigating further I found a way to solve the problem using and interface:
```
var nextCursor interface{} = nil
if len(studies) >= pagination.Number {
  nextCursor = createNextCursor(pagination)
}
```
(I know that interface zero value is nil, but I think it's more useful to make it explicit)

ericrfdev:
@nandanrao Updated the PR with the changes 👍

dashboard-api/internal/platform/server/handler/studies/list.go
```diff
@@ -31,31 +24,17 @@ func ListHandler(studyRepository studiesmanager.StudyRepository) gin.HandlerFunc
...
-				"data": data,
-				"pagination": gin.H{
-					"nextCursor": createNextCursor(pagination),
-				},
-			})
-		}
+		ctx.JSON(http.StatusOK, gin.H{
+			"data": studies,
+			"pagination": gin.H{
+				"nextCursor": nextCursor,
```
nandanrao, line 35: Huh, if this is an interface, how does gin serialize it? Is it using reflection? Curious. Don't we know the type of nextCursor? Isn't it an integer or something?

ericrfdev: It can be either a string or null. I explained in this comment -> https://github.com/vlab-research/vlab/pull/57#discussion_r746160090 why I ended up using an interface.

About gin, it relies on json.Marshal to serialize it, which in turns relies on reflection (```e.reflectValue(reflect.ValueOf(dataToSerialize), opts)```)

nandanrao:
```Go
type Pagination struct {
   NextCursor `json:"nextCursor,omitempty"`
}
```

ericrfdev:
@nandanrao Added a new commit typing explicitly the responses as we discussed in previous meeting

nandanrao:
Hey are you waiting for me to approve this? Looks great!

PR #67, opened 2021-11-28
(#60) Add authorization layer to vlab dashboard api

ericrfdev:
(#60) Add authorization layer to vlab dashboard api
      - Create a middleware to validate JWT tokens
      - Use the middleware in all the endpoints, except the health check one
      - Update endpoints and repositories to be aware of the "userId"
      - Update studies table to have a user_id column (user table will come later, when creating and endpoint to create users)
      - Update health check endpoint to take into account the health of the auth0 service (for us, in this case it means that we can obtain the public key, used to verifiy the JWT tokens by EnsureValidTokenMiddleware)


dashboard-api/internal/platform/server/handler/health/check.go
```diff
@@ -17,22 +18,56 @@ type checkResponseDependencies []struct {
 	Healthy bool   `json:"healthy"`
 }
 
-func CheckHandler(repositories storage.Repositories) gin.HandlerFunc {
+func CheckHandler(repositories storage.Repositories, auth0Domain string) gin.HandlerFunc {
 	return func(ctx *gin.Context) {
-		isHealthy := true
+		cockroachdbHealthyStatus := isCockroachdbdbHealthy(ctx, repositories)
+		auth0HealthyStatus := isAuth0Healthy(ctx, auth0Domain)
```
nandanrao, line 24: Nice touch. Definitely not MVP, but very thoughtful!! ;)

dashboard-api/internal/platform/server/middleware/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+		jose.RS256,
+		josev2.WithExpectedClaims(expectedClaimsFunc),
+	)
+	if err != nil {
+		log.Fatalf("EnsureValidTokenMiddleware: failed to set up the josev2 validator: %v", err)
+	}
+
+	/*
+		This error handler func is required to override the default behaviour of the jwtmiddleware
+		which writes to the headers and body of the response upon error.
```
nandanrao, line 41: A little more explanation - why do we not want to write the error to the body in case of an error? I assume just because we prefer to do the writing ourself?

dashboard-api/internal/platform/server/middleware/auth/auth.go
```diff
@@ -0,0 +1,65 @@
...
+	*/
+	noopErrorHandlerFunc := func(w http.ResponseWriter, r *http.Request, err error) {}
+	middleware := jwtmiddleware.New(validator.ValidateToken, jwtmiddleware.WithErrorHandler(noopErrorHandlerFunc))
+
+	return func(ctx *gin.Context) {
+		var encounteredError = true
+
+		var handler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request) {
```
nandanrao, line 49: I'm getting a bit of bad code smell here, but I don't know the library. 

It feels a bit like they give you a middleware, but you don't use it so you create your own. That's fine. But in that case it's probably more straight-forward to use some lower-level methods, if they give it to you, to just check the JWT. 

In other words, I'd expect something like: 

 ```go
 ok, err :=  validator.ValidateToken(tokenFromCtxSomehow)
if err != nil {
  ctx.AbortWithStatusJSON(...)
}
ctx.Next()
 ```

Would be sufficient. What else is this providing?

nandanrao: (that's psuedo-code, I have no idea what the API is of validator!

nandanrao (commented):
Overall looks good - left one major comment on a part that seems a bit suboptimal.

ericrfdev:
Just to provide some context to anyone viewing, we discussed the PR comments elsewhere.

As a summary we're relying on the go-jwt-middleware to avoid reimplementing the following logic: 
- Fetch the public key from auth0 and caching it in memory for few minutes.

This middleware when attempting to validate the JWT, if it founds and error it writes directly to the response. We wanted to override this behaviour and handle the errors similar to the rest of the api, that's one of the reason why there is some extra boilerplate in the code.

Issue #68, opened 2021-11-29
Add endpoint to create users from a valid JWT token


ericrfdev:
This endpoint is required by the dashbord-client and it's purpose is to sync new users in auth0 to our own database.

The dashboard-client allows visitors to register through auth0. After registration, auth0 sends the user again to the dashboard.

Before displaying anything to the new user, we need to ensure that we have it in our own database.

So the dashboard-client will call this endpoint with the JWT issued from auth0 before displaying anything to the user. 

As an alternative approach we could have relied on auth0 post registration hooks to achieve this. The reasons why we haven't choose this approach are:
- Cannot version control that piece of code alongside the rest of our codebase
- We would have need to use external dead letter queues for non processed events
- What happens if auth0 never triggers the event for certain users?

PR #69, opened 2021-11-29
(#68) Add endpoint to create users from a valid JWT token

ericrfdev:


dashboard-api/internal/platform/server/users/create.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func CreateHandler(repositories storage.Repositories) gin.HandlerFunc {
+	return func(ctx *gin.Context) {
+		user, err := repositories.User.CreateUser(ctx, auth.GetUserIdFrom(ctx))
+
+		if err != nil {
+			switch {
+			case errors.Is(err, studiesmanager.ErrUserAlreadyExists):
+				ctx.JSON(http.StatusUnprocessableEntity, gin.H{"error": "User already exists"})
```
nandanrao, line 24: Cool - so the client will be in charge of knowing that this is chill. Great!

dashboard-api/internal/platform/storage/user_repository.go
```diff
@@ -0,0 +1,35 @@
...
+	return &UserRepository{
+		db: db,
+	}
+}
+
+func (r *UserRepository) CreateUser(ctx context.Context, userId string) (studiesmanager.User, error) {
+	_, err := r.db.Exec("INSERT INTO users (id) VALUES ($1)", userId)
+
+	if err != nil {
+		if err.Error() == "ERROR: duplicate key value violates unique constraint \"primary\" (SQLSTATE 23505)" {
```
nandanrao, line 25: Isn't there a code you can check instead of the entire message? Might be more robust? Any unique constraint violation should suffice here, no?

dashboard-api/internal/user.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+	"errors"
+)
+
+type UserRepository interface {
+	CreateUser(ctx context.Context, userId string) (User, error)
+}
+
+var ErrUserAlreadyExists = errors.New("User Already Exists")
```
nandanrao, line 12: Nice.

nandanrao (approved):
Very straight-forward! Beautiful! One small question, but overall looks great to me, merge away whenever.

dashboard-api/internal/platform/storage/user_repository.go
ericrfdev: Just for context to anyone viewing, we discussed this elsewhere.

We need to identify the errors that happen due the user already exists and provide specific errors that will be used for the clients of the api.

The error code 23505 it's used when any unique constrained it's violated while processing the query. Although right now, the primary key of the table is the only column with a unique constraint, there may be more in the future.

So I decided to go with this approach of comparing the string. It seems flacky but:
- Dependencies are locked
- We'll have metrics and alerts for this endpoint
- Soon there will be e2e that integrates with the real db and will be run as part of our CI/CD pipeline.

If at some point this bite us, we can rethink how we approach it and may change the approach. For example perform an extra query that checks if the user exists before attempting to create it.

PR #71, opened 2021-12-09
(#70) Ensure user exists in our backend before showing the authenticated user any authenticated page

ericrfdev:


dashboard-frontend/src/__tests__/user.spec.tsx
```diff
@@ -46,48 +62,81 @@ describe('user', () => {
...
+      userAuthenticated(true);
+      server.post('/users', () => new Response(500));
+
+      visit('/');
+      screen.getByTestId('loading-page');
+
+      await waitFor(() => {
+        screen.getByText('Oops, something went wrong!');
+        screen.getByText(
+          'Please check your internet connection and try again.'
```
nandanrao, line 76: I would say show the message from the server, and if there is none, just show "500: server error". It's annoying to the user, but that's the backend developers fault for not sending helpful messages!

nandanrao: donde "error.message" es, o del response o del http client.

ericrfdev: Yes makes sense, it will be more useful 👍

This change may require update the layout code for certain resolutions, I'll handle them in the following ticket: https://github.com/vlab-research/vlab/issues/77

Issue #72, opened 2021-12-13
Dockerize dashboard-api in all environments and add a github action to run the tests during CI


ericrfdev:

Spazzy757:
@nandanrao this is complete correct?

nandanrao:
Seems reasonable to me!

PR #75, opened 2022-01-23
(#72) Dockerize dashboard-api in all environments and add a github action to run the tests during CI

ericrfdev:
- Add Dockerfile for PRO environment
- Add bash scripts to start and stop in dev mode all the services that form the 'dashboard-api'
  - Live reloading is supported through air go package
  - The database is seeded with some fixture data
- Add test script and supporting files for bootstrapping the environment and running all the tests
- Add github action that runs the test cases for every new commit pushed to any branch. (a merge to main also count as a push)
  - Note that the action will only be triggered if the commit contains at least one change in a file that is under the dashboard-api folder.

**Note**: I splitted the task in multiple incremental commits to make easier the review 🥳


.github/workflows/dashboard-api-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+  tests:
+    name: Tests
+    runs-on: ubuntu-20.04
+    env:
+      working-directory: ./dashboard-api
+    steps:
+      - uses: actions/checkout@v2
+      - name: Tests
+        run: ./test.sh --is-ci
+        working-directory: ${{env.working-directory}}
```
calufa, line 17: Are we using `working-directory` env var somewhere else besides from line 17? If not, couldn't we just use `./dashboard-api` instead of `${{env.working-directory}}`?

dashboard-api/Dockerfile.test
```diff
@@ -0,0 +1,9 @@
+FROM golang:alpine
+
+RUN apk add --no-cache \
+    gcc \
+    libc-dev \
+    readline-dev
+
+WORKDIR /github.com/vlab-research/vlab/dashboard-api
+RUN go get -u github.com/nathany/looper
```
calufa, line 9: Why do we need a `Dockerfile.test`? Couldn't we just reuse `Dockerfile.dev`?

The idea of having an additional dockerfile apart from `Dockerfile`, is that we wanted to have a container for _development_ purposes, and leave `Dockerfile` with only the strictly necessary dependencies. 

The `Dockerfile.dev` will then contain everything required for local development and testing, as this dockerfile doesn't need to be _optimize_ as `Dockerfile` would.

calufa: In this specific case, it looks like both `looper` and `air` are essentially doing the same thing, right?

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
+services:
+  cockroachdb:
+    image: cockroachdb/cockroach:v21.2.3
```
calufa, line 3: Unless we are running a different version of `cockroach` on this project, at least on `fly` we are using `v20.1.4`, for example, https://github.com/vlab-research/fly/blob/b0e949944028edab871b895da10da49f02fb900e/devops/dev-cluster.sh#L31

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
...
+    entrypoint: cockroach start-single-node --insecure --advertise-addr localhost
+    ports:
+      - "26257:26257"
+      - "8080:8080"
```
calufa, line 7: How are we using port 8080? I thought cockroach only required port 26257?

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
...
+    networks:
+      - dashboard-api
+  setup_cockroachdb:
+    image: cockroachdb/cockroach:v21.2.3
```
calufa, line 11: Unless we are running a different version of cockroach on this project, at least on fly we are using v20.1.4, for example, https://github.com/vlab-research/fly/blob/b0e949944028edab871b895da10da49f02fb900e/devops/dev-cluster.sh#L31

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
...
+    volumes:
+      - ./setup_db.sh:/setup_db.sh
+      - ./seed_cockroachdb.sql:/seed_cockroachdb.sql
```
calufa, line 14: This is another opinion, but I always thought that placing bash scripts at the `root` of the OS was not the best idea in terms of security.

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
...
+    entrypoint: "/bin/bash"
+    command: /setup_db.sh
+    depends_on:
+      - cockroachdb
+    networks:
+      - dashboard-api
+networks:
+  dashboard-api:
```
calufa, line 22: I think we can remove the tailing `:`?

dashboard-api/docker-compose-dev.yaml
```diff
@@ -0,0 +1,17 @@
+services:
+  api:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    environment:
+      STUDIESMANAGER_HOST: 0.0.0.0
```
calufa, line 7: This is a bit strange to me, does using `0.0.0.0` mean that `STUDIESMANAGER` is running inside the same container as the `dashboard-api`?

dashboard-api/docker-compose-common.yaml
```diff
@@ -0,0 +1,22 @@
+services:
```
calufa, line 1: Ideally, we will specify the file version of docker-compose as the first line of this file, for example `version: '3'`, for more info see https://docs.docker.com/compose/compose-file/compose-file-v3/

dashboard-api/docker-compose-dev.yaml
```diff
@@ -0,0 +1,17 @@
+services:
```
calufa, line 1: Ideally, we will specify the file version of docker-compose as the first line of this file, for example `version: '3'`, for more info see https://docs.docker.com/compose/compose-file/compose-file-v3/

dashboard-api/docker-compose-test.yaml
```diff
@@ -0,0 +1,41 @@
+services:
```
calufa, line 1: Ideally, we will specify the file version of docker-compose as the first line of this file, for example `version: '3'`, for more info see https://docs.docker.com/compose/compose-file/compose-file-v3/

dashboard-api/docker-compose-test.yaml
```diff
@@ -0,0 +1,41 @@
...
+  api:
+    build:
+      context: .
+      dockerfile: Dockerfile.test
+    environment:
+      STUDIESMANAGER_DBURI: postgresql://root@cockroachdb:26257/vlab?sslmode=disable
```
calufa, line 7: For consistency reasons, shouldn't we use the same env var declarations everywhere? Currently, in `system` we use https://github.com/vlab-research/vlab/pull/75/files#diff-fc992c9a9d4e6e1728cac26d958fcbe0cde48518b65c36fbd6740aa47d85b8a3R30-R33.

dashboard-api/docker-compose-dev.yaml
```diff
@@ -0,0 +1,17 @@
...
+  api:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
+    environment:
+      STUDIESMANAGER_HOST: 0.0.0.0
+      STUDIESMANAGER_PORT: 80
+      STUDIESMANAGER_DBURI: postgresql://root@cockroachdb:26257/vlab?sslmode=disable
```
calufa, line 9: For consistency reasons, shouldn't we use the same env var declarations everywhere? Currently, in `system` we use https://github.com/vlab-research/vlab/pull/75/files#diff-fc992c9a9d4e6e1728cac26d958fcbe0cde48518b65c36fbd6740aa47d85b8a3R30-R33.

dashboard-api/.air.toml
```diff
@@ -0,0 +1,53 @@
+# Config file for [Air](https://github.com/cosmtrek/air) in TOML format
```
calufa, line 1: Do we really need this file? I'd assume air comes with _safe_ and _sane_ defaults. 
Or could we remove the ones that comes by default?

dashboard-api/seed_cockroachdb.sql
```diff
@@ -16,7 +16,7 @@ VALUES
 
 CREATE TABLE studies(
     id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
-    user_id string NOT NULL REFERENCES users(id), 
+    user_id string NOT NULL REFERENCES users(id) ON DELETE CASCADE,
```
calufa, line 19: I like how enforcing the schema allows you to _discover_ potential future problems early on.

dashboard-api/setup_db.sh
```diff
@@ -0,0 +1,2 @@
+./cockroach sql --host=cockroachdb:26257 --insecure -e "CREATE DATABASE vlab"
```
calufa, line 1: The commands on this script could easily be executed by docker-compose. Moving the content of this script to docker-compose will allow us to have one less file and one less file to mount to the container.

dashboard-api/test.sh
```diff
@@ -0,0 +1,13 @@
...
+docker-compose -f docker-compose-common.yaml -f docker-compose-test.yaml build setup_cockroachdb
+docker-compose -f docker-compose-common.yaml -f docker-compose-test.yaml run setup_cockroachdb
+echo "\n>>> DATABASE IS READY TO RECEIVE CONNECTIONS <<<\n"
+
+echo "\n>>> SETTING UP API <<<\n"
+docker-compose -f docker-compose-common.yaml -f docker-compose-test.yaml build api
+echo "\n>>> FINISHED SETTING UP API <<<\n"
+
+echo "\n>>> RUNNING TEST CASES <<<\n"
+docker-compose -f docker-compose-common.yaml -f docker-compose-test.yaml run api
```
calufa, line 13: If:
- `api` depends on `setup_cockroachdb`, https://github.com/vlab-research/vlab/pull/75/files#diff-fa845e05416b322fd4056e688636b49f1fce19916670dee563a5d479004db617R13
- and `setup_cockroachdb` depends on `cockroachdb`, https://github.com/vlab-research/vlab/pull/75/files#diff-3403c3600e3dae55dc11f60a0884bb3676681159fee1fe90166ad191d2b9da5aR18

Why do we need to _build_ and _run_ `setup_cockroachdb` and `api`? Shouldn't we just _build_ and _run_ `api`?

calufa:
LGTM

I left a few _cosmetic_ comments and _opinions_.

I guess one last comment I have is on where to mount files inside the container. ATM, we are mounting local files to different paths inside the container, like seen here:

- https://github.com/vlab-research/vlab/pull/75/files#diff-fa845e05416b322fd4056e688636b49f1fce19916670dee563a5d479004db617R17
- https://github.com/vlab-research/vlab/pull/75/files#diff-3403c3600e3dae55dc11f60a0884bb3676681159fee1fe90166ad191d2b9da5aR13-R14
- https://github.com/vlab-research/vlab/pull/75/files#diff-fc992c9a9d4e6e1728cac26d958fcbe0cde48518b65c36fbd6740aa47d85b8a3R14

While I don't see this as a _bad_ thing, I personally prefer to put the code, regardless of language or whether it's a bin that I want to run inside the container, in a `/app` folder, mainly for consistency reasons (following the "consistency over correctness" principle). As an `app` for me, it could be anything that is run by a system or user.

A distinction can be made between an `application` and a `program` if religious about definitions, but my preference is to call them `apps`, as an app is a program, but a program doesn't necessarily is an app.

dashboard-api/docker-compose-common.yaml
nandanrao: No need to use the same version. Actually, fly is behind due to an cockroach old bug that has since been fixed, but we need to do the upgrade.

.github/workflows/dashboard-api-tests.yml
ericrfdev: I used the other github action I have as a baseline and I didn't remind to change this. Good catch 👍

dashboard-api/.air.toml
ericrfdev: Yes but as you pointed we can remove most of the options. The ones that are required are cmd and bin.

dashboard-api/Dockerfile.test
ericrfdev: Yes I will group them into a single Dockerfile.dev 👌

dashboard-api/docker-compose-common.yaml
ericrfdev: Makes sense 👍

dashboard-api/docker-compose-common.yaml
ericrfdev: Okey, then I'll leave this as it is :)

dashboard-api/docker-compose-common.yaml
ericrfdev: It exposes the cockroach dashboard that displays some metrics about the database.
I haven't found it useful to troubleshoot any issue, I'll remove this line for now 👌

dashboard-api/docker-compose-common.yaml
ericrfdev: As we discussed in the other comment I'll leave this as it is for now :)

dashboard-api/docker-compose-common.yaml
ericrfdev: In this particular case, I can't come up with an attack vector where having this file at the root can become an issue. Do you had something in mind?

Anyways I ended up inlining the script i the docker-compose file 👌

dashboard-api/docker-compose-common.yaml
ericrfdev: I tried but it seems like it's required

dashboard-api/docker-compose-dev.yaml
ericrfdev: Agree 👍

dashboard-api/docker-compose-dev.yaml
ericrfdev: Yup this name is confusing, actually the STUDIESMANAGER is the DASHBOARD_API.

I changed the naming of the env variables to be more aligned with the names I used for the system service. 👌

dashboard-api/docker-compose-dev.yaml
ericrfdev: Agree, will update the names to be more aligned 👍

dashboard-api/test.sh
ericrfdev: Omg my bad 🤯  good catch !!

dashboard-api/docker-compose-common.yaml
calufa: IMO, scripts and data (_malicious_) in `/` can access all files in `/` recursively (including `bin`, `lib`, `sys`, `tmp`), and perform actions on those files, such as deleting files, etc.

Unless you are _sensitive_ and _vigilant_ about folder and file permissions in `/`, changing OS-dependent dirs (using `chmod`), could create OS-level side effects. 

On the other hand, if you place all app files in `/{DIR_NAME}` and set the dir permissions appropriately, you get a logical _sandbox_ for all files in the app. The app then, will be _self-contained_ and living in a single place, without any other file or dir that is not _relevant_ to the it.

dashboard-api/docker-compose-dev.yaml
calufa: Ok, in that case. Do you think we should change the directory name and other occurrences of `dashboard-api` to match the env var name `studies-manager`? 

Having two different ways of naming the same thing could cause confusion. My preference is to be consistent with naming.

dashboard-api/docker-compose-dev.yaml
calufa: Sorry for the confusion, but what I meant to say is that instead of declaring a single URI string for the DB, we split it into multiple env vars. 

Instead of this:
```
STUDIESMANAGER_DBURI: postgresql://root@cockroachdb:26257/vlab?sslmode=disable
```

we do this:
```
DATABASE_NAME: vlab
DATABASE_HOST: cockroachdb
DATABASE_PORT: 26257
DATABASE_USER: root
```

This will be more consistent with how we define DBs in `system`, see https://github.com/vlab-research/vlab/pull/75/files#diff-fc992c9a9d4e6e1728cac26d958fcbe0cde48518b65c36fbd6740aa47d85b8a3R30-R33.

dashboard-api/docker-compose-common.yaml
ericrfdev: Yes you're right, thanks for the examples 👍

calufa (approved):
Looks great!

PR #95, opened 2022-07-14
Feature/authentication tab

sssoja:


dashboard-frontend/src/pages/ConnectPage/ConnectedAccountsPage.tsx
```diff
@@ -0,0 +1,19 @@
+import PageLayout from '../../components/PageLayout';
+import ConnectedAcccountsList from './ConnectAccountsList';
+
+const accounts = [
+  { id: '1', name: 'Fly', slug: 'fly', authType: 'secret' },
```
nandanrao, line 5: I like it. Probably don't need a separate "id" - the name is unique, no?

dashboard-frontend/src/pages/ConnectPage/ConnectAccountsList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  account: { name, slug, authType },
+}: {
+  account: AccountResource;
+}) => (
+  <li>
+    <div className="px-4 py-4 sm:px-6">
+      <div className="flex flex-col sm:flex-row sm:items-center">
+        <p className="text-sm font-medium text-indigo-600 truncate">{name}</p>
+
+        {authType === 'secret' ? (
```
nandanrao, line 31: It'd be a lot easier to read if these were just separate (if similar) components!

nandanrao:
Looks good!! What should the backend look like? What about mocking it out with mirage?

dashboard-frontend/src/pages/ConnectPage/ConnectedAccountsPage.tsx
sssoja: added `id` back in because TyepScript was throwing an error, I think because Mirage automatically adds an id property to the response.

sssoja:
Authentication tab @nandanrao 

- When logged in and the user navigates to `/accounts` they should see a list of accounts that are universal to all users
- The list of accounts to connect to will rarely change so this can be coded as a regular component using React
- Instead use Mirage to seed and fetch account data for each user
- We can imagine the data to include properties like `name, authType, email, credentials: { clientID, clientSecret, token}, createdAt` etc
- If credential data exists for the user for a given account, the UI should indicate to the user that they have already connected
- If already connected, the user should be able to update their credentials
- If no credential data exists for the user for a given account, the UI should indicate to the user that they have yet to connect

dashboard-frontend/src/helpers/arrays.ts
```diff
@@ -1 +1,28 @@
...
+  }
+  return arr1;
+};
+
+const matcher = (el: any, arr2: [], key: string) => {
+  for (let i = 0; i < arr2.length; i++) {
+    if (el[key] === arr2[i][key]) {
+      return arr2[i];
+    }
+    if (!el[key] === arr2[i][key]) {
```
nandanrao, line 23: What's the opposite of el[key]? I'm a bit confused by this one.

dashboard-frontend/src/helpers/arrays.spec.ts
```diff
@@ -11,3 +11,61 @@ describe('lastValue', () => {
...
+      { name: 'foo', data: 'abcd' },
+      { name: 'bar' },
+      { name: 'baz', data: '1234' },
+    ];
+    const res = arrayMerge(arr1, arr2, 'name');
+
+    expect(res).toStrictEqual(expectation);
+  });
+
+  it('merges the properties of three matching array elements into another', () => {
```
nandanrao, line 54: I don't know if three is really needed here!

dashboard-frontend/src/helpers/arrays.ts
sssoja: Good point, not sure why I thought the second `if` statement was necessary...

sssoja: You're right! I realised this didn't make sense and noticed that it was actually exiting the loop at line 20 and running the condition again which gave me the desired result for the wrong reason at line 23. 

I've now refactored the if statement to reflect this.

PR #98, opened 2022-09-08
Feature/auth tab

sssoja:


dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
```diff
@@ -0,0 +1,184 @@
+import React, { useState } from 'react';
+import { AccountResource } from '../../types/account';
+import { arrayMerge } from '../../helpers/arrays';
+import PrimaryButton from '../../components/PrimaryButton';
+import InputSecret from './InputSecret';
+import InputToken from './InputToken';
+import useCreateAccount from './useCreateAccount';
+import useUpdateAccount from './useUpdateAccount';
+
+const staticAccountResources = [
```
Spazzy757, line 10: Im assuming these are all the account types we support?

dashboard-frontend/src/pages/AccountsPage/InputToken.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ChangeEventHandler } from 'react';
+import { classNames } from '../../helpers/strings';
+
+const InputToken = ({
```
Spazzy757, line 4: I noticed that we show these in plain text, which is fine for a first iteration but in general secret values like tokens and client secrets we should set the input type to password (potentially with the ability to unhide them)

Spazzy757:
Looks good! I'll approve for now but set a label "blocked" until we get the dashboard API bit out or else this will break the frontend

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
sssoja: Exactly, except test!

dashboard-frontend/src/pages/AccountsPage/InputToken.tsx
sssoja: Sounds reasonable! created an issue for it https://github.com/vlab-research/vlab/issues/111

PR #102, opened 2022-10-12
Feature/create study config

sssoja:
- created configs for all form types
- created controllers for all config types 
- built out the React components to render different forms dynamically based on config type 
- added unit testing for the following controllers: `simple` / `select` – todo `list`


sssoja:
Closing to work on feat: create study with a general conf #127

Issue #105, opened 2023-02-21
Chore: Update Authentication Middleware


Spazzy757:
While setting up the Dashboard API locally It complains that  `josev2` is not part of the the `auth0/go-jwt-middleware` package anymore.

```bash
github.com/vlab-research/vlab/dashboard-api/internal/platform/server/middleware/auth imports
	github.com/auth0/go-jwt-middleware/validate/josev2: module github.com/auth0/go-jwt-middleware@latest found (v1.0.1), but does not contain package github.com/auth0/go-jwt-middleware/validate/josev2
```

could we rewrite the `EnsureValidTokenMiddleware` to not need the Jose library at all (potentially just validating the claims and signature of the JWT token)

PR #106, opened 2023-02-22
Chore: Update Auth0

Spazzy757:
# Changelog

- `github.com/auth0/go-jwt-middleware/validate/josev2` has been removed from the auth0 library in favour of `github.com/auth0/go-jwt-middleware/v2/validator`
- updated local dev environment (I added a Makefile in order to make things slightly easier to get started with)
- added documentation to help with getting started and development
- added unit testing for the Auth Middleware

FIX: https://github.com/vlab-research/vlab/issues/105


Spazzy757:
@nandanrao All tests seem to pass and it seems to run without issue. Let me know your thoughts on structure and docs

dashboard-api/internal/platform/server/middleware/auth/auth_test.go
```diff
@@ -0,0 +1,144 @@
...
+		KeyID:     "kid",
+		Algorithm: string(jose.RS256),
+		Use:       "sig",
+	}
+}
+
+// SetupTestJWKServer creates a fake server to respond to JWK requests
+// TODO: Move to internal/testhelpers, currently not possible due to cycle
+// import
+func SetupTestJWKServer(
```
nandanrao, line 80: This seems like a great thing to have for tests, makes it all very easy to test fully!

dashboard-api/Makefile
```diff
@@ -0,0 +1,60 @@
+default: help
```
nandanrao, line 1: Love having a Makefile. 

Seems like there are two versions, one where we use docker compose and one where we just run `go test`. 

The docker compose, I assume, is for running the DB? But `go test` doesn't need the DB, or...?

dashboard-api/internal/platform/server/middleware/auth/auth_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+	"github.com/go-playground/assert/v2"
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+func TestEnsureValidTokenMiddleware(t *testing.T) {
```
nandanrao, line 17: Very reasonable!!

dashboard-api/internal/platform/server/middleware/auth/auth_test.go
```diff
@@ -0,0 +1,144 @@
...
+	audience := "aud"
+	jwk := GenerateJWK(t)
+	testServer := SetupTestJWKServer(t, jwk)
+	defer testServer.Close()
+
+	t.Run("invalid token", func(t *testing.T) {
```
nandanrao, line 23: I would make the name a bit more descriptive so we know _why_ the token is invalid (i.e. the audience is wrong)

dashboard-api/Makefile
Spazzy757: Yes this was some confusion for me as well as you have a `docker-compose.test.yaml`, but you don't actually need a test. I added the target as it was in the test.sh file

dashboard-api/internal/platform/server/middleware/auth/auth_test.go
Spazzy757: :100: will update it

dashboard-api/Makefile
Spazzy757: Let me know if you want me to delete the docker-compose stuff

dashboard-api/internal/platform/server/middleware/auth/auth_test.go
Spazzy757: This was taken from Auth0's library so cant take all the credit

dashboard-api/Makefile
nandanrao: Yeah, if it's not being used, clean it out!

Spazzy757: Did a clean up of unnecessary files as well.

.github/workflows/dashboard-api-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Dashboard Api Tests
+name: Dashboard Api
```
Spazzy757, line 1: Changed this because its redundant, i.e it will now be `Dashboard Api / Tests` instead of `Dashboard Api Tests / Test`

Spazzy757:
Also @nandanrao I removed looper (its been archived) Im not sure why we were using it (looks like the tests reran every time you made changes), but you can do hot reloads using air (I've updated the config to watch go files) Let me know if this fits

nandanrao:
Sounds good. I think looper was there to do TDD - it's what I always use locally - really just an alternative for `watch go test`. Although, to be honest, not sure what advantage is offers. 

But ok - it all looks good to me!

Issue #110, opened 2023-03-08
Feat: Backend API For Connected Accounts


Spazzy757:
## Situation

We currently have a Pull Request open for connected accounts tab on the dashboard. It makes use of an `/api/accounts` endpoint. This does not currently exist, so this ticket will be used to track progress of that work. 

As far as I can see we need to do the following work:

- [x] Create Accounts Repo to Handle database operations (Get, Create, Update, Delete) for Accounts
- [x] Endpoint to Create Accounts (Post)
- [ ] Endpoint To Get/List Accounts (Get)
- [ ] Endpoint to Update Accounts (Put)
- [ ] Endpoint to Delete Accounts

PR #112, opened 2023-03-08
feat: adds api for account creation

Spazzy757:
# Changelog

- API Endpoint to Create Accounts

Part of the work for https://github.com/vlab-research/vlab/issues/110


dashboard-api/internal/platform/storage/account_repository.go
```diff
@@ -0,0 +1,61 @@
...
+	ctx context.Context,
+	a studiesmanager.Account,
+) (studiesmanager.Account, error) {
+
+	c, err := json.Marshal(a.ConnectedAccount.Credentials)
+	if err != nil {
+		return studiesmanager.Account{}, handleCreateError(err, a)
+	}
+
+	q := "INSERT INTO credentials (user_id, entity, key, details) VALUES ($1, $2, $3, $4)"
```
Spazzy757, line 34: @nandanrao just want to confirm that "Accounts" in the frontend are actually "Credentials" in the database

Spazzy757:
The tests are failing because of the requirement to have the database running in the CI, need to merge https://github.com/vlab-research/vlab/pull/85

dashboard-api/internal/platform/server/handler/accounts/create.go
```diff
@@ -0,0 +1,82 @@
...
+	if err := validateCredentials(a); err != nil {
+		return &studiesmanager.Account{}, err
+	}
+
+	return a, nil
+}
+
+// validateCredentials is a simple validator used to handle the unstructured
+// nature of account credentials
+func validateCredentials(a *studiesmanager.Account) error {
```
nandanrao, line 61: If we are anyways having _some_ of the structure and validating, I guess I'm not convinced it isn't just as easy to have _all_ the structure and use Go Types, which is also easy to read and can be shared elsewhere, no? 

i.e. given we're all in one Go module, we have a struct called `FlyCredentials` already in the Fly connector. Why not just use that? We can just try to marshal the json into all the supported types and if it fails, throw an error.

Spazzy757: Yeah happy to do this! Was thinking it might be better, we could also potentially use generics for the creation then :thinking:

dashboard-api/internal/account.go
```diff
@@ -0,0 +1,85 @@
...
+	AccountBase
+	Credentials FlyCredentials `json:"credentials" validate:"required"`
+}
+
+// TODO: we could use the type in `inference/sources/fly/main.go`
+// for this, however we would need to add the validator to it
+// this also might break "domain" interfaces and be hard to decouple later
+type FlyCredentials struct {
+	APIKey string `json:"api_key" validate:"required"`
+}
```
Spazzy757, line 56: @nandanrao let me know your thoughts on this

dashboard-api/internal/account.go
```diff
@@ -0,0 +1,85 @@
...
+
+type AccountBase struct {
+	ID        string    `json:"id" `
+	UserID    string    `json:"userId"`
+	AuthType  string    `json:"authType" validate:"required"`
+	Name      string    `json:"name" validate:"required"`
+	CreatedAt time.Time `json:"createdAt"`
+}
+
+func (a AccountBase) GetName() string {
```
nandanrao, line 34: Are these getters really necessary? Feels Java.

dashboard-api/internal/account.go
nandanrao: Adding validator tags seems harmless to me!

dashboard-api/internal/account.go
Spazzy757: Yes, as the type will change based on i.e `Typeform` account and/or `Fly` account so we use these on the interface in order to maintain type integrity (i.e we know whatever we pass into the create function will definitely have this and wont cause a nil pointer)

dashboard-api/internal/account.go
```diff
@@ -0,0 +1,100 @@
...
+// logic in order to determine if an account has been "connected".
+// if this is null, then no account is connected
+type FlyConnectedAccount struct {
+	CreatedAt   time.Time      `json:"createdAt"`
+	Credentials FlyCredentials `json:"credentials" validate:"required"`
+}
+
+// TODO: we could use the type in `inference/sources/fly/main.go`
+// however that is an executable and not a package so we would need
+// to refactor that code
```
Spazzy757, line 60: @nandanrao I can create a PR to refactor the fly source so we can use it here

dashboard-api/internal/account.go
Spazzy757: For reference, I started this work and then came across that that source is actually a executable and not a package so we need to refactor it:
```
internal/account.go:9:2: import "github.com/vlab-research/vlab/inference/sources/fly" is a program, not an importable package
```

dashboard-api/internal/account.go
Spazzy757: Im actually completley wrong, I changed it form using embeded structs and just used an interface on the `ConnectedAccount` which seems to remove the need for getters

PR #114, opened 2023-03-13
feat: move fly credentials into package so its reusable

Spazzy757:
# Changelog

- We would like to reuse the `FlyCredentials` struct in the dashboard API (see PR: https://github.com/vlab-research/vlab/pull/112) however you can not import main packages in go, so this PR moves the `FlyCredentials` to a sub-package `types` in order that we can update that PR to use it


nandanrao (approved):
LGTM!

PR #115, opened 2023-03-13
chore: setup docker build for dashboard api

Spazzy757:
# Changelog

- adds a docker build to the Github actions for the dashboard API

## Notes on how this works 
This will do two things.

On Pull Requests it will create one image:

- `vlabresearch/vlab-dashboard-api:{head_commit_sha}` .

This is just to validate that a PR does not break the build process. 

When you tag it will create an image with two tags:
-  `vlabresearch/vlab-dashboard-api:{head_commit_sha}`
- `vlabresearch/vlab-dashboard-api:latest`

This also changes the pipeline that it will ALWAYS build on main branch changes, this is generally handy in monorepos that share dependencies across services (i.e shared golang packages) so that any changes will be validated and you don't run into the problem where a change in one package breaks your app, but you don't pick it up until deploying to production.


Spazzy757:
image


The CI/CD seems to work well, I enabled it only to push on master so we dont get rate limits from branches

Issue #119, opened 2023-03-15
Feat: Design API to handle Study Configuration Creation


Spazzy757:
## Situation

With all the great work @sssoja and @nandanrao have done for a Study Configuration Flow on The Dashboard UI, we now need a backend API to manage that configuration. 

This API Should conform to the structure laid out in the Frontend PR: https://github.com/vlab-research/vlab/pull/118

We should break the logic down into two sections, noteably:

- [x] The API Layer that accepts the data from the frontend
- [x] A translation Layer that translates the configuration into what the database is expecting.

PR #120, opened 2023-03-15
feat: studies config api

Spazzy757:
# Changelog

- Creates and API endpoint to create a study configuration under a study, this accepts the data from the frontend and transforms it into what the database holds

Fix: https://github.com/vlab-research/vlab/issues/119


devops/sql/init.sql
```diff
@@ -18,7 +18,7 @@ CREATE TABLE studies(
 
 CREATE TABLE study_confs(
        created TIMESTAMP DEFAULT CURRENT_TIMESTAMP(),
-       study_id UUID NOT NULL REFERENCES studies(id),
+       study_id UUID NOT NULL REFERENCES studies(id) ON DELETE CASCADE,
```
Spazzy757, line 21: @nandanrao im assuming if a study is deleted we want to cleanup any related config.

dashboard-api/internal/platform/server/handler/studyconfig/create_test.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func TestHandler_StudyConfiguration_Create(t *testing.T) {
+	assert := require.New(t)
+	userId := "auth0|61916c1dab79c900713936de"
+	studyslug := "test-study"
+	testcases := []struct {
+		studyconfig    types.StudyConfig
+		studyslug      string
+		description    string
```
sssoja, line 21: should description not fall under the `studyconfig` struct?

dashboard-api/internal/types/studyconfig.go
```diff
@@ -0,0 +1,125 @@
...
+	Gender   string `json:"gender"`
+	Location string `json:"location"`
+}
+
+// DatabaseStudyConf is the structure used to store data
+// in the database
+type DatabaseStudyConf struct {
+	CreatedAt time.Time
+	StudyID   string
+	ConfType  string
```
sssoja, line 123: So `ConfType` here refers to "General", "Targeting" etc?

In the frontend `config.type` is the form type rather than the kind of configuration that we're creating a study with, maybe I could do some renaming in the FE to change `config.type` to `form.type` to make that distinction? I don't know if it really matter though.

dashboard-api/internal/types/studyconfig_test.go
```diff
@@ -0,0 +1,94 @@
...
+				Conf:     []byte(`{"age":"21","gender":"F","location":"Spain"}`),
+			},
+		}
+		s, err := input.TransformForDatabase()
+		assert.NoError(err)
+		for i, _ := range expected {
+			assert.Equal(expected[i], s[i])
+		}
+	})
+	t.Run("does not transform null configs", func(t *testing.T) {
```
sssoja, line 57: Cool! Only thing I don't grasp is how the "null" test is different to the happy path except for `assert.Equal(len(s), 2)` can you show me what that means?

dashboard-api/internal/platform/server/handler/studyconfig/create_test.go
Spazzy757: So this is something called table testing, its just a really fancy way of simplifying your tests into basic structures. The description here is reffering to the test description and not the StudyConfig. Theres a small blog about it [here](https://dev.to/boncheff/table-driven-unit-tests-in-go-407b)

dashboard-api/internal/types/studyconfig.go
Spazzy757: Yes exactly! but I would say this shouldnt matter to the frontend as the backends job is to worry about this and do the mappings

dashboard-api/internal/types/studyconfig_test.go
Spazzy757: Yes lets go through it tomorrow morning?

dashboard-api/internal/platform/server/handler/studyconfig/create_test.go
sssoja: gotcha

dashboard-api/internal/platform/server/server.go
```diff
@@ -55,13 +56,18 @@ func (s *Server) registerRoutes() {
...
 
+	//APIS used for the studies resource
 	s.Engine.GET("/studies/:slug", s.ensureValidTokenMiddleware, studies.ReadHandler(s.repositories))
 	s.Engine.GET("/studies", s.ensureValidTokenMiddleware, studies.ListHandler(s.repositories))
 	s.Engine.POST("/studies", s.ensureValidTokenMiddleware, studies.CreateHandler(s.repositories))
-
 	s.Engine.GET("/studies/:slug/segments-progress", s.ensureValidTokenMiddleware, segmentsprogress.ListHandler(s.repositories))
 
+	//APIS for the study configurations resource
+	s.Engine.POST("/studies/:slug/configuration", s.ensureValidTokenMiddleware, studyconfig.CreateHandler(s.repositories))
```
nandanrao, line 69: While I appreciate the specificity, given that the types and tables everywhere are called "confs", it seems almost confusing to switch to a more proper word.

dashboard-api/internal/types/studyconfig.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+type StudyConfigRepository interface {
+	Create(ctx context.Context, dsc DatabaseStudyConf) error
+}
+
+// StudyConfig is the root config that holds the various
+// different types of study configuration that can be
+// applied to a study that gets sent by the frontend
+type StudyConfig struct {
```
nandanrao, line 25: The database expects each config separately, which leads to an odd struct iteration when "transforming for database" - and the database version, `[]DatabaseStudyConf`, is an array and therefore doesn't even validate that they're all there. And one can imagine updating one without updating the others. 

To add to that, one can imagine that a user wants to "partially create" a study, create and save SOME of the confs, without saving all of them. 

Given all of that, I'm not sure it's right for the frontend to send `StudyConfig` as this large object, rather than just send each config individually. 

I think also if we think of validation as an first-class, but potentially asynchronous, process. It might be nice to have a `studies/:id/validate` endpoint that validates all the current configuration and returns any issues. As a simple MVP it miht just validate that it has all the confs.

devops/sql/init.sql
nandanrao: Yeah... Need to come up with migrations at some point...

dashboard-api/internal/platform/server/server.go
Spazzy757: Will change it to StudyConf

dashboard-api/internal/types/studyconfig.go
Spazzy757: I have added an issue for the validation endpoint https://github.com/vlab-research/vlab/issues/122 and made the discussed changes for now.  I will work on the patch once this is done.

Issue #124, opened 2023-03-21
Feat: Get StudyConf endpoint


Spazzy757:
## Situation

We have merged the creation of a study configuration, we now need and endpoint that will fetch the latest Study COnfig to display in the UI. As this is an Append only table (we dont validate or check config when creating, we just add latest config), we need to filter out the latest config for every config type and create our configuration structure from this.

PR #125, opened 2023-03-21
feat: add get study config endpoint

Spazzy757:
# Changelog

- adds get study config endpoint

Fixes: https://github.com/vlab-research/vlab/issues/124

### DIFF
--- a/dashboard-api/internal/platform/server/handler/studyconf/read.go
+++ b/dashboard-api/internal/platform/server/handler/studyconf/read.go
@@ -1,0 +1,57 @@
+package studyconf
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	studiesmanager "github.com/vlab-research/vlab/dashboard-api/internal"
+	"github.com/vlab-research/vlab/dashboard-api/internal/platform/server/middleware/auth"
+	"github.com/vlab-research/vlab/dashboard-api/internal/platform/storage"
+	"github.com/vlab-research/vlab/dashboard-api/internal/types"
+)
+
+type readResponse struct {
+	Data types.StudyConf `json:"data"`
+}
+
+// ReadHandler is a gin handler used to retrieve a studies configuration
+func ReadHandler(r storage.Repositories) gin.HandlerFunc {
+	return func(ctx *gin.Context) {
+		var req struct {
+			Slug string `uri:"slug" binding:"required"`
+		}
+		if err := ctx.ShouldBindUri(&req); err != nil {
+			ctx.Status(http.StatusBadRequest)
+			return
+		}
+
+		dscs, err := r.StudyConf.GetByStudySlug(ctx, req.Slug, auth.GetUserIdFrom(ctx))
+		if err != nil {
+			handleError(ctx, err)
+			return
+		}
+
+		sc := types.StudyConf{}
+		err = sc.TransformFromDatabase(dscs)
+		if err != nil {
+			handleError(ctx, err)
+			return
+		}
+
+		ctx.JSON(http.StatusOK, readResponse{Data: sc})
+	}
+}
+
+func handleError(ctx *gin.Context, err error) {
+	switch {
+	case errors.Is(err, studiesmanager.ErrStudyNotFound):
+		ctx.JSON(http.StatusNotFound, gin.H{"error": err.Error()})
+		return
+	default:
+		msg := fmt.Sprintf("an unexpected error occured - %v", err.Error())
+		ctx.JSON(http.StatusBadRequest, gin.H{"error": msg})
+		return
+	}
+}
--- a/dashboard-api/internal/platform/server/handler/studyconf/read_test.go
+++ b/dashboard-api/internal/platform/server/handler/studyconf/read_test.go
@@ -1,0 +1,71 @@
+package studyconf_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"github.com/vlab-research/vlab/dashboard-api/internal/platform/storage"
+	"github.com/vlab-research/vlab/dashboard-api/internal/testhelpers"
+	"github.com/vlab-research/vlab/dashboard-api/internal/types"
+)
+
+func TestHandler_StudyConfiguration_GetByStudySlug(t *testing.T) {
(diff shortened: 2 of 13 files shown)


dashboard-api/internal/platform/server/handler/studyconf/read.go
```diff
@@ -0,0 +1,62 @@
...
+			return
+		}
+
+		s, err := r.Study.GetStudyBySlug(ctx, req.Slug, auth.GetUserIdFrom(ctx))
+		if err != nil {
+			handleError(ctx, err)
+			return
+		}
+
+		dscs, err := r.StudyConf.GetByStudyID(ctx, s.Id)
```
nandanrao, line 36: Why not just create a GetByStudySlug (and user)? Seems more efficient to just put that join in the query, no?

dashboard-api/internal/types/studyconf.go
```diff
@@ -46,28 +49,50 @@ func (sc *StudyConf) TransformForDatabase() (res []DatabaseStudyConf, err error)
...
 }
 
+// TransformFromDatabase gets a list of DatabaseStudyConf and iterates
+// through it, adding the relevant configs to the StudyConf struct
+func (sc *StudyConf) TransformFromDatabase(dcs []*DatabaseStudyConf) error {
+	for _, d := range dcs {
+		//TODO: Im sure we can make this more future proof by using
+		//reflect and checking the json tags
+		switch d.ConfType {
+		case "general":
```
nandanrao, line 79: I would make a map just to make it slightly cleaner, as we'll keep adding, and the logic will be identical.

nandanrao:
Left some comments, nothing blocking. Overall, I approve!

dashboard-api/internal/platform/server/handler/studyconf/read_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func TestHandler_StudyConfiguration_GetByStudySlug(t *testing.T) {
+	assert := require.New(t)
+	id := uuid.New()
+	studyslug := id.String()
+	testcases := []struct {
+		databasestudyconfs []*types.DatabaseStudyConf
+		studyslug          string
+		description        string
```
sssoja, line 22: Not a feedback just a comment that this is the struct I was referring to earlier when I was redundantly passing the `slug` and the `description` from the frontend!

My understanding now is that the client is simply sending the json data that exists inside the `databasestudyconfs` and that this struct is just a concept for storing response data in the backend

dashboard-api/internal/testhelpers/testhelpers.go
```diff
@@ -136,12 +137,20 @@ func DeleteAllUsers() {
...
 }
 
 func CreateUser() {
 	r := GetRepositories()
 	_, _ = r.Db.Exec("INSERT INTO users (id) VALUES ($1)", CurrentUserId)
 }
+
+func CreateDatabaseStudyConf(dsc types.DatabaseStudyConf) error {
+
+	r := GetRepositories()
```
sssoja, line 152: excuse my silly Go question: is that `r` for respositories and `q` for query?

dashboard-api/internal/platform/server/handler/studyconf/read.go
Spazzy757: True, I can update that

dashboard-api/internal/platform/server/handler/studyconf/read_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"github.com/vlab-research/vlab/dashboard-api/internal/platform/storage"
+	"github.com/vlab-research/vlab/dashboard-api/internal/testhelpers"
+	"github.com/vlab-research/vlab/dashboard-api/internal/types"
+)
+
+func TestHandler_StudyConfiguration_GetByStudySlug(t *testing.T) {
```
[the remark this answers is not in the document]
Spazzy757: Maybe we can discuss this in stand up, im not sure I understand what you mean :sweat_smile: There is only one test for the Get endpoint which does end-to-end

dashboard-api/internal/platform/server/handler/studyconf/read_test.go
Spazzy757: Ah right, no this structure is specifically what the database expects and not what the forntend would pass, its the backends job to convert what the frontend passes into this

dashboard-api/internal/testhelpers/testhelpers.go
Spazzy757: yes, this is definitely not one of my favorite practices in Golang, but they tend to push towards very simple variables names that understanding can be pulled from the context of the decleration, in this case:
```golang
// We can assume that r is now a repositories structure due to the naming of the function
r  :=  GetRepositories()
```
There is quite a large [styleguide](https://google.github.io/styleguide/go/best-practices#variable-declarations) for golang written by google:

dashboard-api/internal/types/studyconf.go
```diff
@@ -46,28 +49,50 @@ func (sc *StudyConf) TransformForDatabase() (res []DatabaseStudyConf, err error)
 		// service use to identify the config
 		fieldName := v.Type().Field(i).Tag.Get("json")
 		// if a field is not set we dont need to add it to our array
-		// as it will just have empty data
-		if field.IsZero() {
+		// as it will just have empty data as well as if it is not one of
+		// the confs which should all be pointers
+		if field.IsZero() || field.Kind() != reflect.Ptr {
```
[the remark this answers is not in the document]
Spazzy757: so `Kind()` is essentially the type of the structure so in this case we are checking if its a type `Pointer` (you can also have types like `string` or `int` etc)

dashboard-api/internal/platform/storage/studyconf_repository.go
```diff
@@ -43,3 +43,37 @@ func (r *StudyConfRepository) Create(
 
 	return nil
 }
+
+// GetConfsByStudy will return all the latest unique confs that are attached to
+// a study
+func (r *StudyConfRepository) GetByStudyID(
```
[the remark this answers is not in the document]
Spazzy757: I think this is what Nandan is mentioning earlier. In the database slug is not unique so but it makes sens that it should be unqiue

PR #127, opened 2023-03-23
feat: create study with a general conf

sssoja:
- implemented a simple controller that takes a conf of type `confObject` and returns a set of fields
- developed a `formBuilder` that takes a single field and returns a translated version configured for rendering a form
- separated the concept of creating a new study and creating a new study conf
- user can create a new study at the `/studies` endpoint
- user can post study conf data to the /studies/:slug/conf (in prog)


dashboard-frontend/src/pages/StudyConfPage/configs/general.js
```diff
@@ -0,0 +1,107 @@
+export const general = {
+  type: 'confObject',
+  title: 'General',
+  description: 'The "general" configuration consists of... General stuff?',
+  fields: [
```
Spazzy757, line 5: Do we know what the actual values we could use for these are?

i.e maybe we can check our current data to find out what possible values we could have for:
- `optimization_goals`
- `objective`

Spazzy757:
Im happy with how this looks and functions, Im not the best to ask about code styling for Typescript but it looks like easy to follow code

dashboard-frontend/mocks/state.js
```diff
@@ -0,0 +1,286 @@
...
+    ],
+    general: [
+      {
+        id: 'objective',
+        name: 'objective',
+        type: 'select',
+        component: Select,
+        label: 'Objective',
+        helper_text: undefined,
+        options: [
```
nandanrao, line 26: I don't know if we need to support _all_ the options for objectives and optimization_goals, but this is great for now!

dashboard-frontend/src/helpers/api.ts
```diff
@@ -96,6 +94,32 @@ const createStudy = ({
...
+  accessToken,
+}: {
+  data: StudyConfResource;
+  slug: string;
+  accessToken: string;
+}) =>
+  apiRequest<CreateStudyConfApiResponse>(`/studies/${slug}/conf`, {
+    accessToken,
+    method: 'POST',
+    body: { ...data },
```
nandanrao, line 109: `body: data`, no?

dashboard-frontend/src/helpers/arrays.ts
```diff
@@ -1 +1,8 @@
 export const lastValue = <T>(values: T[]): T => values[values.length - 1];
+
+export const reducer = (arr: any[]) => {
```
nandanrao, line 3: this should have a more informative name!

dashboard-frontend/src/helpers/formBuilder.ts
```diff
@@ -0,0 +1,57 @@
+import Text from '../pages/NewStudyPage/components/form/inputs/Text';
+import Select from '../pages/NewStudyPage/components/form/inputs/Select';
+import { ConfBase, FieldBase } from '../types/form';
+import { createNameFor } from './strings';
+
+const str: keyof FieldBase = 'type';
+
+export const formBuilder = (field: FieldBase, localFormData?: any) => {
```
nandanrao, line 8: I don't love the thing called `formBuilder` builds a field, not a form... Another naming issue ;)

dashboard-frontend/src/helpers/getField.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+      name: 'instagram_id',
+      type: 'text',
+      component: Text,
+      label: 'Instagram Id',
+      helper_text: 'E.g 2327764173962588',
+      options: undefined,
+      value: '',
+    };
+
+    const res = fields.length > 1 && getField(fields, event);
```
nandanrao, line 27: `fields.length > 1` seems like an odd thing here, maybe no longer needed?

dashboard-frontend/src/hooks/useStudyConf.tsx
```diff
@@ -0,0 +1,26 @@
+import { useQuery, queryCache } from 'react-query';
+import useAuthenticatedApi from './useAuthenticatedApi';
+
+const useStudyConf = (slug: string) => {
+  const useStudyConfQuery = (slug: string) => {
```
nandanrao, line 5: Not sure I understand why this is defined as a function

dashboard-frontend/src/pages/NewStudyPage/components/form/Form.tsx
```diff
@@ -0,0 +1,116 @@
...
+  } = useCreateStudyConf();
+
+  const [fieldState, setFieldState] = useState<FieldState[]>();
+  const [globalFormData, setFormGlobalFormData] = useState<
+    StudyConfResource | any
+  >({
+    fetchedConfData,
+  });
+
+  const localFormData = !isCreateStudyForm && fetchedConfData[title];
```
nandanrao, line 34: I don't like `isCreateStudyForm` as a prop. Why is there distinct logic needed for that form?

dashboard-frontend/src/pages/NewStudyPage/components/form/Form.tsx
```diff
@@ -0,0 +1,116 @@
...
+import Fieldset from './Fieldset';
+import { reducer } from '../../../../helpers/arrays';
+import PrimaryButton from '../../../../components/PrimaryButton';
+import useCreateStudyConf from '../../../../hooks/useCreateStudyConf';
+import { useParams } from 'react-router-dom';
+import { StudyConfResource } from '../../../../types/study';
+
+export const Form = (props: any) => {
+  const { conf, controller, isCreateStudyForm, data } = props;
+  const fetchedConfData = data?.data.data;
```
nandanrao, line 14: data data data?

dashboard-frontend/src/pages/NewStudyPage/controllers/simple.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Conf } from '../../../types/conf';
+import { ConfBase, FieldState } from '../../../types/form';
+
+const simple = (
+  conf: ConfBase,
+  localFormData?: Conf,
+  event?: Event,
+  fieldState?: FieldState[]
+) => {
+  const type: string = 'confObject';
```
nandanrao, line 16: Why is this needed? I thought it was the other way around. `confObject` just declares that it uses this controller, why does the controller know about the object type?

dashboard-frontend/src/pages/NewStudyPage/controllers/simple.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!localFormData && !fieldState && !event) {
+      return initialiseFieldState(conf);
+    }
+
+    if (!localFormData && fieldState && event) {
+      return updateFieldState(fieldState, event);
+    }
+
+    if (localFormData && !fieldState && !event) {
+      return getFieldState(conf, localFormData);
```
nandanrao, line 28: I think we had this conversation, so forgive me if I'm forgetting something, but I'm not convinced localFormData is needed here at all. Why can't fieldState contain everything, including current values? 

This refers to Form as well. I'm not sure why localFormData is needed as a concept. It seems to me we have `fieldState`, which contains the fields and their values. Then, we have a translation to "formData" type.

nandanrao (changes requested):

dashboard-frontend/src/helpers/api.ts
Spazzy757: No that will make it:
```json
{
  "data": {
    "genera: {
    }
  }
}
```
When what we want is:
```json
{
  "general": {
    ...
  }
}
```

So I think this is correct

dashboard-frontend/src/helpers/arrays.ts
Spazzy757: agree with @nandanrao here, hard to understand what this function is reducing. Can we make an issue to rename it though ad I dont see it as a blocked to merge

sssoja: Updated to `reduceFieldStateToAnObject` and made the test description more informative too.

dashboard-frontend/src/helpers/formBuilder.ts
sssoja: Renamed to `translateField`, I think that makes more sense

dashboard-frontend/src/helpers/getField.spec.ts
sssoja: Yes, old code! removed

dashboard-frontend/src/hooks/useStudyConf.tsx
sssoja: Fair point, just following previous code patterns, have refactored to do away with the function and set `studyConfQuery` as a variable with a value of `useQuery(['studyConf', slug], () =>
    fetchStudyConf({ slug }));`

dashboard-frontend/src/pages/NewStudyPage/components/form/Form.tsx
sssoja: It was purely to distinguish between which react query hook to invoke on submit, if it's a create study form, run `createStudy()`, if a conf form, run `createStudyConf()`. 

You're right though, the prop isn't needed. I've refactored so that there's no longer a prop and the form simply makes a check at the `handleSubmit` to see if current `title` is `create_study`, based on whatever the boolean is it knows which hook to run.

dashboard-frontend/src/pages/NewStudyPage/components/form/Form.tsx
sssoja: Fixed!

dashboard-frontend/src/pages/NewStudyPage/controllers/simple.ts
sssoja: No that's valid, there's no reason for it, it was just an extra guardrail in case any other conf type slipped in but I don't see why this would happen.

dashboard-frontend/src/pages/NewStudyPage/controllers/simple.ts
sssoja: So when we fetch the existing form data, where should we plug those values into the value prop of each field object?

Does it not make sense to do that in the `translateField` helper like this? 

```
return {
    id: field.name,
    name: field.name,
    type: type,
    component: component,
    label: field.label,
    helper_text: field.helper_text ?? field.helper_text,
    options: field.options?.map((option: ConfBase | any) =>
      option.title
        ? {
            name: createNameFor(option.title),
            label: option.title,
          }
        : option
    ),
    value: localFormData ? localFormData[field.name] : getInitialValue(field),
  };
```
For this to work though we would need `localFormData` as a concept. I see what you mean about `fieldState` being the only data store needed to hold the field values but I think i'm missing something obvious here!

dashboard-frontend/src/helpers/api.ts
sssoja: Actually, this is turning out to be tricky! I tried spreading the data obj at the component level before passing it to the api call and I wasn't able to get this to return a response without using the spread operator on the `body`

Spazzy757 (approved):

PR #129, opened 2023-03-23
feat: add creatives to study conf

Spazzy757:
# Changelog

- adds creative conf to the study conf setup to allow the API to start setting it


Spazzy757:
We need to wait for https://github.com/vlab-research/vlab/pull/125 to be merged before merging this PR

Spazzy757:
> I'm worried that we're going to have test/mock data littered everywhere of the different conf types, making it hard to change them.
> 
> Any thoughts on centralizing test/mock data in one place so it's easy to update?

I was thinking of writing a helper but then I starting going down the rabbit hole of how to make a helper generic enough to allow for different data. My thoughts where once we have all the conf implemented some kind of pattern would arise that we could abstract.

Spazzy757:
> I'm worried that we're going to have test/mock data littered everywhere of the different conf types, making it hard to change them.
> 
> Any thoughts on centralizing test/mock data in one place so it's easy to update?

See: https://github.com/vlab-research/vlab/pull/136

PR #142, opened 2023-03-28
feat: add organisation

Spazzy757:
# Changelog

- adds the idea of an organisation so that multiple users can work on the same study
- Removes all mocking except for in storage where we use sql-mock
- refactor Id -> ID (this is based on Golangs intilism concept that dictates that Initlisms or abreviations should always be Uppercase see the [golang wiki](https://github.com/golang/go/wiki/CodeReviewComments#initialisms))


Spazzy757:
We need to wait for: https://github.com/vlab-research/vlab/pull/144 to be merged

dashboard-api/internal/platform/server/handler/studies/create.go
```diff
@@ -21,18 +21,24 @@ type createResponse struct {
...
 // a new Study object in the database
-func CreateHandler(repositories storage.Repositories) gin.HandlerFunc {
+func CreateHandler(r storage.Repositories) gin.HandlerFunc {
 	return func(ctx *gin.Context) {
 		var req, err = parseRequest(ctx)
 		if err != nil {
 			ctx.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 			return
 		}
+		orgID, err := r.User.GetUserOrgID(ctx, auth.GetUserIdFrom(ctx))
```
nandanrao, line 31: Walk me through your logic on this. I would expect this to either be in middleware OR purely in the queries (join users), as we discussed, but this approach seems to me to have the worst of both worlds: we're touching all the queries AND have a separate DB query AND have to put it into every handler manually? Or is the thought to put it into a middleware later as a refactor?

Spazzy757: Ah right I complely misunderstood our conversation, this is why I was against this approach :sweat_smile:

dashboard-api/internal/types/user.go
```diff
@@ -7,10 +7,17 @@ import (
 
 type UserRepository interface {
 	Create(ctx context.Context, userID string) (User, error)
+	GetUserOrgIDs(ctx context.Context, userID string) ([]Org, error)
```
nandanrao, line 10: `GetUserOrgs` - it doesn't just get the IDs.

Spazzy757: Yes this is left over from when I was just retuning ids

dashboard-api/internal/storage/user.go
```diff
@@ -20,17 +19,70 @@ func NewUserRepository(db *sql.DB) *UserRepository {
...
 	}
 
 	return types.User{
-		ID: userID,
+		ID: id,
 	}, nil
 }
+
+// GetUserOrgID returns a users org id based on their user id
+func (r *UserRepository) GetUserOrgIDs(
```
nandanrao, line 62: `GetUserOrgs`

dashboard-api/internal/storage/user.go
```diff
@@ -20,17 +19,70 @@ func NewUserRepository(db *sql.DB) *UserRepository {
 }
 
 func (r *UserRepository) Create(ctx context.Context, userID string) (types.User, error) {
-	_, err := r.db.Exec("INSERT INTO users (id) VALUES ($1)", userID)
+	q := `
```
nandanrao, line 22: This seems reasonable to me, but I can't figure out where it's tested, am I not looking in the right place? Seems like a reasonable thing to have a couple simple user tests around, especially if one wants to refactor the query in the future (i.e return org id, etc.).

Spazzy757: I haven't tested this yet 😅 I'm still in the processes of figuring out approach (I mentioned this o. Slack)

dashboard-api/internal/server/middleware/organizations/organizations.go
```diff
@@ -0,0 +1,27 @@
+package organizations
+
+import (
+	"github.com/gin-gonic/gin"
+)
+
+type orgHeader struct {
+	orgID string `header:"VLAB_ORGANIZATION"`
```
nandanrao, line 8: Cool, so the idea is that all requests should, in addition to a user, have org in the header? 

I'm not sure I love it. It feels a little non-traditional to me. It feels unclear if this is authorization or something else. 

I'm trying to work through the need. We have a couple of resources, right now: 

1. Studies. A study belongs to a single org.

I look at Vercel here, just because I've been using it recently, and notice that the projects are underneath teams in the URL: /{TEAM}/{PROJECT}. Which, in a resource/REST scenario, is easy to understand and makes sense to me. Studies belong to Orgs. 

2. Accounts / Connected Accounts. When one connects an external account, one inherently does so as an individual. But one connects the external account to... What? The Org? Or all Orgs the user is a part of? From an external perspective, it doesn't need to be the Org - once Facebook authorizes Vlab to act on behalf of an individual, Vlab has that permissionsfro period. Regardless of Org. From on internal perspective, I might connect my Facebook account to one Org, but not want to connect it to another Org. That seems reasonable to me. But then maybe there are two separate concepts: (a) did Nandan authorize Vlab to use Facebook or (b) did Nandan connect his Facebook to Org X. 

So one could either (1) subordinate account to an org or (2) separate those two concepts, subordinate (b) to the Org but leave (a) at an individual level. 

Thoughts?

PR #144, opened 2023-03-29
chore: refactor study segments

Spazzy757:
# Changelog

- the study segments handler did not follow the codebases patterns therefore it made it difficult to add features due to the inconsistancies.
- removed the study segment mocks that are not needed
- Changed the get segment progress endpoint to make one database call instead of two


Spazzy757:
This is just a refactor  in order to make the organisations work started on PR: https://github.com/vlab-research/vlab/pull/142 easier to validate

PR #145, opened 2023-03-30
feat: add destinations to study conf

Spazzy757:
# Changelog

- adds destination to study conf api for dashboard

### DIFF
--- a/dashboard-api/internal/platform/server/handler/studyconf/read_test.go
+++ b/dashboard-api/internal/platform/server/handler/studyconf/read_test.go
@@ -29,18 +29,21 @@ func TestHandler_StudyConfiguration_GetByStudySlug(t *testing.T) {
 			},
 			expectedStatus: 200,
 			studyslug:      studyslug,
-			expectedRes:    `{"data":{"general":{"name":"Foo","objective":"","optimization_goal":"link_clicks","destination_type":"Web","page_id":"1","min_budget":1,"opt_window":48,"instagram_id":"","ad_account":"12345"},"targeting":null,"targeting_distribution":null,"recruitment":null}}`,
-			description:    "return 200 for valid studyconfig account with only general",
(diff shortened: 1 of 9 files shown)


dashboard-api/internal/platform/server/handler/studyconf/read_test.go
```diff
@@ -58,7 +61,7 @@ func TestHandler_StudyConfiguration_GetByStudySlug(t *testing.T) {
 					assert.NoError(err)
 				}
 				res := getStudyConfRequest(t, tc.studyslug)
-				assert.Equal(res.Body, tc.expectedRes)
+				assert.Equal(tc.expectedRes, res.Body)
```
nandanrao, line 64: classic.

dashboard-api/internal/types/studyconf.go
```diff
@@ -172,6 +177,88 @@ type RecruitmentConf struct {
...
+		return err
+	}
+
+	dests := []Destination{}
+
+	for _, jd := range jsonDest {
+		switch {
+		//WebDestination should always have URLTemplate
+		//set
+		case jd.URLTemplate != "":
```
nandanrao, line 209: Interesting. Duck typing in Go. I like it. 

But I think after you've ducktyped, determined what it is, you should just re-marshal the JSON into the correct type, no? 

Another option -> use `gjson` to avoid marshaling until you know what it is.

nandanrao: This smells funky to me... I can't really understand what's going on, tbh. 

It feels like we're marshalling it into a Go Type, but then putting it into a weird interface that's anyways useless, so we're throwing away all type information. At that point, what's the point? Why not just leave it as rawJson?

nandanrao (changes requested):

dashboard-api/internal/types/studyconf.go
Spazzy757: So this structure allows for us to have a list of different types (This is the object of the interface), so when we add the validation we can specify required fields etc based on the types (The types are still there). I agree with the GJson but keeping them in raw JSON will make it hard for validation top occur.

PR #154, opened 2023-04-03
feat: create recruitment conf

sssoja:
# Changelog

- created a single select form that renders a dropdown where the user can select from 3 recruitment types
- the user can see previously configured recruitment data
- known issue #162


Spazzy757:
Just for my understanding, for destination Recruitment:

image

Will we be adding the list of destinations once we have destinations in the conf?

Spazzy757:
I did a functional review, all seems to be working perfectly, Im happy to merge this :+1:

PR #155, opened 2023-04-04
chore: fixes routepath routing

sssoja:
# Changelog

The route path was showing a blank screen on page load – this fixes that by taking the user directly to /studies


Spazzy757 (approved):
LGTM

PR #157, opened 2023-04-05
Feat connected accounts tab

Spazzy757:
# Changelog

This is a squashed branch of https://github.com/vlab-research/vlab/pull/98

- Creates a tab for connected accounts to be managed in the UI


dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
```diff
@@ -0,0 +1,175 @@
...
+  // to the account
+  function handleCredentialChange(e: any): void {
+    setCredential({
+      ...credential,
+      [e.target.name]: e.target.value,
+    });
+  }
+
+  // a bit of duck typing to return the correct input box
+  // TODO this seems complicated, we shouldnt need it
```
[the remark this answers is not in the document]
sssoja: As the only difference here is the construct of the key meaning both should take a prop of `key` and once passed into a component of `Input` you can then separate into two concepts

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
```diff
@@ -0,0 +1,175 @@
...
+        handleCredentialChange={handleCredentialChange}
+        index={index}
+        entity={account.authType}
+        flyApiKey={credential.api_key}
+      />
+    ),
+  };
+
+  const validateCredentials = JSON.parse(
+    JSON.stringify(credential),
```
sssoja, line 77: Unless we want to intentionally return "undefined", we can just say does a value exist? if true give me the value otherwise give me null. 

The shorthand way of writing this would be `value ?? value`

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  return (
+    <li data-testid="account-list-item">
+      <div className="px-4 py-4 sm:px-6 py-6">
+        <div className="flex flex-col sm:grid sm:grid-cols-5 sm:gap-4">
+          <h2 className="mb-4 text-sm font-medium text-indigo-600 truncate sm:mb-0 sm:col-span-1">
+            {account.name}
+          </h2>
+          <form onSubmit={handleSubmitForm} className="col-span-3">
+            {inputType[account.name] || undefined}
```
sssoja, line 106: Same as above as it seems we're only checking for existence rather than intentionally wanting to return undefined so what about this: 

```
inputType[account.name] && {
 <div className="flex items-baseline">
              <PrimaryButton
                type="submit"
                testId={`existing-account-submit-button-${index}`}
                loading={isCreating}
              >
                {account.connectedAccount &&
                account.connectedAccount.credentials
                  ? 'Update'
                  : 'Connect'}
              </PrimaryButton>
            </div>
          </form>
        </div>
      </div>
}
``` 

This is basically saying, if this thing exists, render this, otherwise do nothing.

dashboard-frontend/src/pages/AccountsPage/AccountsPage.tsx
```diff
@@ -0,0 +1,99 @@
...
+  setOpen,
+}: {
+  open: boolean;
+  setOpen: React.Dispatch<React.SetStateAction<boolean>>;
+}) => {
+  // createAccounts is used to hold a list of accounts that has not been 
+  // added to the database, but instead is in an "intermediary" phase
+  const [createAccounts, setCreateAccounts] = useState<AccountResource[]>([]);
+
+  const addAccountHandler = (a: AccountResource) => {
```
sssoja, line 35: rename `a` to `account`

dashboard-frontend/src/pages/AccountsPage/InputFlyAPIKey.tsx
```diff
@@ -0,0 +1,62 @@
...
+        id={`${index}`}
+        name="identifier"
+        type="text"
+        className={classNames(
+          'block w-full shadow-sm sm:text-sm rounded-md',
+          (error ?? '').trim() !== ''
+            ? 'focus:ring-red-500 focus:border-red-500 border-red-300 text-red-900 pr-10'
+            : 'focus:ring-indigo-500 focus:border-indigo-500 border-gray-300'
+        )}
+        placeholder="Name of Credential"
```
sssoja, line 32: credential name?

sssoja: or give your credential a name

dashboard-frontend/src/pages/AccountsPage/InputTypeformKey.tsx
```diff
@@ -0,0 +1,61 @@
...
+        placeholder="Enter API Key"
+        onChange={handleCredentialChange}
+        value={typeformkey}
+        required
+      />
+    </div>
+  </React.Fragment>
+);
+
+export default InputTypeformKey;
```
sssoja, line 61: Nice that these two components are so similar as they can definitely be dried out at a later date into one!

dashboard-frontend/src/types/account.ts
```diff
@@ -0,0 +1,31 @@
...
+  extends ApiResponse<AccountResource> {}
+
+export interface AccountResource {
+  id?: string;
+  name: string;
+  authType: string;
+  connectedAccount?: ConnectedAccountResource;
+}
+
+export interface ConnectedAccountResource {
```
sssoja, line 16: I think we can drop Resource from our naming now. This was something Eric implemented but I think it's a little wordy. Connected Account etc still makes sense to me

sssoja (approved):

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { useState } from 'react';
+import { AccountResource } from '../../types/account';
+import PrimaryButton from '../../components/PrimaryButton';
+import InputTypeformKey from './InputTypeformKey';
+import InputFlyAPIKey from './InputFlyAPIKey';
+import useCreateAccount from './useCreateAccount';
+
+const AccountList: React.FC<any> = ({
```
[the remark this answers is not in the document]
Spazzy757: Im not sure I understand :thinking: I thought in typescript you should type everything so that the compiler knows the types its dealing with? Although I might be wrong

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
Spazzy757: This is left over :sweat_smile: I didnt write this, but I can update that

dashboard-frontend/src/pages/AccountsPage/AccountCreateModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+            >
+              <Dialog.Panel className="relative transform overflow-hidden rounded-lg bg-white text-left shadow-xl transition-all sm:my-8 sm:w-full sm:max-w-lg">
+                <div className="bg-white px-4 pb-4 pt-5 sm:p-6 sm:pb-4">
+                  <div className="sm:flex sm:items-start">
+                    <div className="mt-3 text-center sm:ml-4 sm:mt-0 sm:text-left">
+                      <Dialog.Title
+                        as="h3"
+                        className="text-base font-semibold leading-6 text-gray-900"
+                      >
+                        Add New Connected Account
```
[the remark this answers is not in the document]
Spazzy757: Changed to `New Connected Account` to match what is on the studies page

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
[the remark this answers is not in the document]
sssoja: That's my understanding too tbh! let's reverse what I said, probably the right thing to do is to add `React.FC` to all other components rather than leave it out here

dashboard-frontend/src/pages/AccountsPage/AccountList.tsx
Spazzy757: Not sure if im missing something but the above doesnt seem to work, it only shows the button and not the input

dashboard-frontend/src/pages/AccountsPage/InputTypeformKey.tsx
Spazzy757: Yeah I really dont like it but kept it :sweat_smile:

Spazzy757: Actually solved this

dashboard-frontend/src/pages/AccountsPage/InputAccountCredentials.tsx
```diff
@@ -0,0 +1,129 @@
...
+  name,
+  value,
+  error,
+  handleCredentialChange,
+  index,
+}) => {
+  return (
+    <div className="flex flex-col mb-4">
+      <label className="block mb-2 text-sm font-medium text-gray-900 dark:text-gray-300">
+        {startCase(name)}
```
sssoja, line 107: I was unsure why the lodash lib was needed here but makes sense as I guess it covers keys for both types of credentials e.g 1233foobar and super_secret. It goes a step further than what we have in the `createLabelFor` function in the strings helper file. 

Would we be better off using Lodash for all string manipulation do you think?

dashboard-frontend/src/pages/AccountsPage/InputAccountCredentials.tsx
```diff
@@ -0,0 +1,129 @@
...
+const InputAccountCredentials: React.FC<inputAccountCredentials> = ({
+  error,
+  handleCredentialChange,
+  index,
+  entity,
+  credentials,
+}) => (
+  <React.Fragment>
+    <NameInput name={entity} index={index} />
+    <CredentialsInputList
```
sssoja, line 22: Yep, I like it!

sssoja: The only thing is readability, I'd move as many of those React F.Cs as possible into their own files and import them in.

So `NameInput`, `CredentialsInputList`, and `CredentialsInput` can all be abstracted away. As it will increase the number of files within the `AccountsPage` folder it might also be worth creating a `components` folder where you can move stuff like this. 

For study confs for example I have a structure like this `components/inputs`

sssoja: That's just my personal preference though, do what makes sense to you!

sssoja (approved):

dashboard-frontend/src/pages/AccountsPage/InputAccountCredentials.tsx
sssoja: On second thoughts, I'd probably still keep `CredentialsInputList` and `CredentialsInput` together as they're closely tied and it's pretty easy to see that one is just a list version of the other

dashboard-frontend/src/pages/AccountsPage/InputAccountCredentials.tsx
Spazzy757: Lodash is pretty widely used from my understanding, the only downside is that it's huge so we need to be careful to just import what we need

PR #160, opened 2023-04-05
feat: add audiences to study conf api

Spazzy757:
# Changelog

- adds audiences to the study conf API


dashboard-api/internal/types/studyconf.go
```diff
@@ -266,6 +270,48 @@ type CreativeConf struct {
...
+}
+
+type QuestionTargeting struct {
+	OP   string      `json:"op"`
+	Vars []TargetVar `json"vars"`
+}
+
+type TargetVar struct {
+	Type  string `json:"type"`
+	Value string `json:"value"`
```
Spazzy757, line 290: @nandanrao I see in adopt this is:
```python
class TargetVar(BaseModel):
    type: str
    value: Union[str, int, float]
```
However in the database we only seem to have strings for values, is this acceptable or should I just make TargetVars an blank interface?

nandanrao: We should leave it as an interface/rawJson because while most times we use a string, there's no reason we couldn't use a number.

dashboard-api/internal/types/studyconf.go
```diff
@@ -266,6 +270,48 @@ type CreativeConf struct {
...
+type AudienceConf struct {
+	Name              string             `json:"name"`
+	SubType           string             `json:"subtype"`
+	Lookalike         *Lookalike         `json:"lookalike,omitempty"`
+	QuestionTargeting *QuestionTargeting `json:"question_targeting"`
+	Partioning        *Partioning        `json:"partioning,omitempty"`
+}
+
+type QuestionTargeting struct {
+	OP   string      `json:"op"`
```
nandanrao, line 284: It's short "Operation" so I wouldn't capitalize both letters, no?

PR #165, opened 2023-04-10
fix: change min budget to be a float32

Spazzy757:
# Changelog

- changes data type for min budget to be a float32 (we don't use float64 as float32 is more than big enough to hold monetery amounts)


sssoja (approved):
LGTM! I'm assuming then that no validation needs to be made in the frontend with regard to whether the number is an integer or a decimal, both would be accepted values here

Spazzy757:
> LGTM! I'm assuming then that no validation needs to be made in the frontend with regard to whether the number is an integer or a decimal, both would be accepted values here

Thats correct, javascript will handle this without you needing to

PR #192, opened 2023-04-24
feat: added alchemer source connector

nandanrao:


nandanrao:
@Spazzy757 - Wanted to make a PR so you saw adding a new connector - would love your eyes on this and any thoughts.

inference/sources/alchemer/main.go
```diff
@@ -0,0 +1,337 @@
...
+	return e.Message
+}
+
+func (c *AlchemerConnector) loadEnv() *AlchemerConnector {
+	err := env.Parse(c)
+	handle(err)
+	return c
+}
+
+type AlchemerCreds struct {
```
Spazzy757, line 144: Could we put this in its own package as the api backend will need to use it for the connected accounts

Spazzy757: I moved around Flys one into a package called `types`

Spazzy757:
Looks good, but do we not need to add it to a manifest somewhere in order to deploy it?

Spazzy757 (approved):
Will address the changes in my API changes when I add Alchemy to connected accounts

PR #193, opened 2023-04-24
chore: rename api and dashboard

Spazzy757:
# Changelog

- renames `dashboard-api` => `api`
- renames `dashboard-frontend` => `dashboard`
- updates README
- update Github Workflows


Spazzy757:
> Should we change the Docker container name as well?

I would say we only do this when we start looking at automating deployments

PR #196, opened 2023-04-25
chore: add alchemer to api and ui

Spazzy757:
# Changelog

- created a shared types package in inference sources
- added alchmer to the CI/CD
- adds alchemer to the UI and API


nandanrao:
This looks good, but I'm not convinced that adding a new type should result in new tests, if there's no new business logic in the system related to the type. That makes me wonder a bit.

Spazzy757:
> This looks good, but I'm not convinced that adding a new type should result in new tests, if there's no new business logic in the system related to the type. That makes me wonder a bit.

All honesty it doesnt but I like to have the new case there as a "just in case" as we do duck typing and it just clarifies that our duck typing works for this type as well

PR #206, opened 2023-05-01
feature: general study conf 

sssoja:
# Changelog

- implemented forms for creating a study, general and recruitment confs using the react hook form lib 
- the user can create a study
- the user can add, save and edit data for general and recruitment confs


api/internal/server/handler/facebook/token.go
```diff
@@ -35,7 +35,7 @@ func GenerateToken(a *fb.App, r storage.Repositories) gin.HandlerFunc {
 		}
 		a := types.Account{
 			UserID:   auth.GetUserIdFrom(ctx),
-			AuthType: "bearer",
```
Spazzy757, line 38: :laughing: looks like you have some merge issues

dashboard/src/package.json
Spazzy757, line 1: Is this file necessary?

recruitment/.DS_Store
Spazzy757, line 1: Need to delete this one

dashboard/src/package.json
Spazzy757, line 1: Think this one can be deleted

sssoja: No I just checked and it's identical to the package.json in the root –  removed

recruitment/.DS_Store
sssoja: I've tried but this is proving tricky. I wasn't able to locate the file locally to delete it so I tried rebasing from today's latest main branch. In doing so I've been unable to get rid of this message: 
```
The following untracked working tree files would be overwritten by merge:
	recruitment/.DS_Store
Please move or remove them before you merge.
```
Full context here:

Screenshot 2023-05-05 at 13 43 31

Spazzy757 (approved):
LGTM The only thing I picked up is that the destinations recruitment type wont really be valid until we have destinations and in and the ability to choose one of our destinations

PR #214, opened 2023-05-29
feature: destinations conf

sssoja:
# Changelog

- implemented the destinations form 
- user can input, add, save, edit and delete destinations


Spazzy757 (approved):
The changes look good, but your commit history has diverged so you cant merge this, we can go through it in tomorrows session

Spazzy757 (approved):

PR #229, opened 2023-07-27
refactor: abstract input components

sssoja:


dashboard/src/pages/StudyConfPage/forms/general/General.tsx
```diff
@@ -134,21 +82,45 @@ const General: React.FC<Props> = ({ id, localData, confKeys }: Props) => {
...
+    e.preventDefault();
+
     const data = {
       [id]: formData,
     };
 
     createStudyConf({ data, studySlug });
   };
 
+  const validateInput = (name: string, value: any) => {
```
nandanrao, line 95: Hm. Can we use type information to do this? Is there a more general solution to this problem?

sssoja: I see what you mean so create some kind of validator that says if this field is of type `number` run this validation otherwise give me the raw value.

